PR #3, opened 2021-03-15
Riley/c 2379 call graphql endpoint to get inbox data

rileylnapier:
## Description

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/hooks/use-inbox.ts
```diff
@@ -0,0 +1,15 @@
...
+  string, 
+  { config: IInboxConfig, clientKey?: string}
+] = () => {
+  const {inboxConfig, clientKey} = useCourier<{
+    inboxConfig: IInboxConfig
+  }>();
+
+  const inbox = "inbox";
+
+  return [inbox, { config: inboxConfig ?? {}, clientKey }];
```
rileylnapier, line 14: to do, what should the hook do?  all i really needed was config when i wrote this

packages/react-provider/babel.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
         extensions: [".css"],
       },
     ],
-    ["babel-plugin-react-remove-properties", { properties: ["data-test-id"] }],
+    ["babel-plugin-react-remove-properties", { properties: ["data-testid"] }],
```
rileylnapier, line 13: testid seemed to be the default so we don't need a file to tell it the new pattern

packages/react-provider/src/index.tsx
```diff
@@ -11,26 +15,64 @@ export const CourierContext = React.createContext<ICourierContext | undefined>(
...
+enum ActionType {
+  INIT_TOAST = "INIT_TOAST",
+  INIT_INBOX = "INIT_INBOX",
+}
+interface IAction {
+  type: ActionType;
+  payload: any;
+}
+
+const reducer = (state, action) => {
```
rileylnapier, line 27: not sure im super into the reducer pattern just yet... but its working

RoyalSix (approved):

PR #4, opened 2021-03-17
get components to talk to eachother

rileylnapier:
## Description

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/components/Inbox/index.tsx
+++ b/packages/react-inbox/src/components/Inbox/index.tsx
@@ -1,12 +1,16 @@
-import React, { useState, useEffect } from "react";
+import React, { useEffect } from "react";
 import Messages from "../Messages";
 import Tippy, { TippyProps } from "@tippyjs/react";
 import tippyCss from "tippy.js/dist/tippy.css";
 import styled, { ThemeProvider, createGlobalStyle } from "styled-components";
 import Bell from "./Bell";
-import { useCourier, useActions } from "@trycourier/react-provider";
+import { useCourier } from "@trycourier/react-provider";
+
+import useInbox from "~/hooks/use-inbox";
 
 import { InboxProps } from "../../types";
+import reducer from "~/reducer";
+
 const GlobalStyle = createGlobalStyle`${tippyCss}`;
 
 const StyledTippy = styled(Tippy)(({ theme }) => ({
@@ -39,34 +43,27 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
     throw new Error("Missing Courier Provider");
   }
 
-  const courierActions = useActions();
-  const [visible, setVisible] = useState(false);
-
-  const handleBellOnClick = () => {
-    setVisible(!visible);
-  };
+  const inbox = useInbox();
 
   let tippyProps: TippyProps = {
-    placement: "right",
+    trigger: props.trigger ?? "click",
+    placement: props.placement ?? "right",
     interactive: true,
   };
 
-  if (visible) {
-    tippyProps = {
-      ...tippyProps,
-      visible,
-    };
-  }
+  useEffect(() => {
+    courierContext.registerReducer("inbox", reducer);
+  }, []);
 
   useEffect(() => {
-    if (!courierActions) {
-      return;
-    }
-
-    courierActions.initInbox({
-      inboxConfig: props.config,
+    inbox.init({
+      config: props.config,
     });
   }, [props]);
+
+  if (!courierContext?.reducers?.inbox) {
+    return null;
+  }
 
   return (
     <ThemeProvider theme={props.theme ?? {}}>
@@ -75,7 +72,7 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
         {props.renderIcon ? (
           <span>{props.renderIcon({})}</span>
         ) : (
-          <Bell className={props.className} onClick={handleBellOnClick} />
+          <Bell className={props.className} />
         )}
       </StyledTippy>
     </ThemeProvider>
--- a/packages/react-inbox/src/components/Messages/index.tsx
+++ b/packages/react-inbox/src/components/Messages/index.tsx
@@ -4,7 +4,8 @@ import Message from "../Message";
 import { InboxProps } from "../../types";
 import { Body, Footer, Header, Loading } from "./styled";
 
-import useMessages from "../../hooks/use-messages";
+import useInbox from "../../hooks/use-inbox";
+import useMessages from "~/hooks/use-messages";
 
 const Messages: React.FunctionComponent<InboxProps> = ({
   title = "Inbox",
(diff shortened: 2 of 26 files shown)


packages/react-inbox/src/reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+      };
+    }
+
+    case "inbox/NEW_MESSAGE": {
+      return {
+        ...state,
+        messages: [
+          {
+            created: new Date().getTime(),
+            messageId: action.payload.messageId ?? new Date().getTime(),
```
rileylnapier, line 37: we don't pass messageId through the websocket but we need some kind of id for react key so using datetime for now

suhasdeshpande: what happens if there is any action/side-effect associated with this? Say mark read or delete? We would need an id to do anything with this right?

packages/react-provider/src/transports/base.ts
```diff
@@ -2,25 +2,49 @@ import { ICourierEvent, Interceptor } from "./types";
...
   protected emit = (courierEvent: ICourierEvent): void => {
-    if (!this.listener) {
-      console.warn("No Listener Registered");
+    if (!this.listeners.length) {
+      console.warn("No Listeners Registered");
       return;
     }
-    this.listener(courierEvent);
+
+    for (const listener of this.listeners) {
```
suhasdeshpande, line 23:
```suggestion
    for (const {listener} of this.listeners) {
```

packages/react-provider/src/transports/base.ts
```diff
@@ -2,25 +2,49 @@ import { ICourierEvent, Interceptor } from "./types";
...
+      listener.listener(courierEvent);
+    }
```
suhasdeshpande, line 25:
```suggestion
    for (const { listener } of this.listeners) {
      listener(courierEvent);
    }
```

packages/react-provider/src/transports/base.ts
```diff
@@ -2,25 +2,49 @@ import { ICourierEvent, Interceptor } from "./types";
...
 
   /** Setter method for a listener */
-  listen = (listener: (courierEvent: ICourierEvent) => void): void => {
-    this.listener = listener;
+  listen = (listener: {
+    id: string,
+    listener: (courierEvent: ICourierEvent) => void
+  }): void => {
+    let didReplaceListener = false;
+    this.listeners.map(l => {
```
suhasdeshpande, line 34: I would recommend forEach maybe if we are only going to be looping it for updating `didReplaceListener`

packages/react-provider/src/types.ts
```diff
@@ -1,9 +1,20 @@
...
+}
 export interface ICourierContext {
   apiUrl?: string;
   clientKey?: string;
   dispatch?: (mergeContext: any) => void;
+  reducers?: {
+    [key: string]: any;
+  };
+  registerReducer?: (
+    scope: string,
```
suhasdeshpande, line 15: can we have a more restrictive type here instead of `string`?

packages/react-toast/src/hooks/index.ts
```diff
@@ -20,19 +20,22 @@ export const useListenForTransportEvent = (clientKey: string, transport: ICourie
...
 
-      if (clientKey && courierData?.deliveredUrl) {
-        fetch(`${courierData?.deliveredUrl}`, {
-          method: "POST",
-          headers: {
-            [COURIER_CLIENT_HEADER]: clientKey,
-          },
-        });
-      }
+        if (clientKey && courierData?.deliveredUrl) {
```
suhasdeshpande, line 28: seems like auto-formatter?

packages/react-inbox/src/reducer.ts
rileylnapier: yeah, the way it works right now is that we have tracking links in action.payload.data that will fire off... but that isn't all hooked up yet

packages/react-provider/src/transports/base.ts
rileylnapier: i am using the new array.  there could be a new "listener" function that we need to apply

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -39,43 +43,36 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
-      ...tippyProps,
-      visible,
-    };
-  }
-
   useEffect(() => {
-    if (!courierActions) {
-      return;
-    }
+    courierContext.registerReducer("inbox", reducer);
```
RoyalSix, line 55: As much as I wish there was another way to do this. I actually like this pattern 👍🏾 
It's explicit and its abstracted.
Although I am wondering if there any race conditions with the `inbox.init`. I see the 
```
if (!courierContext?.reducers?.inbox) {
```
below

packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -4,15 +4,17 @@ import Message from "../Message";
...
+import useMessages from "~/hooks/use-messages";
 
 const Messages: React.FunctionComponent<InboxProps> = ({
   title = "Inbox",
   renderHeader,
   renderFooter,
   renderMessage,
 }) => {
-  const messageResults = useMessages();
+  useMessages();
```
RoyalSix, line 16: Does this need to be called on every render?

RoyalSix: Or if its not returning anything maybe it can located somewhere else

packages/react-provider/src/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    case "INIT_INBOX": {
+      return {
+        ...state,
+        inboxConfig: action.payload.config,
+      };
+    }
+
+    case "GET_MESSAGES": {
```
RoyalSix, line 56: Could this state live in the toast reducer if the Toast component does not need to know about messages?

packages/react-provider/src/transports/base.ts
RoyalSix: Wouldn't this be using the new array?
```
const newArray = this.listeners.map(l => l)
```

packages/react-inbox/src/components/Inbox/index.tsx
rileylnapier: there actually might be another way... ive been thinking about it and might play with it today

packages/react-inbox/src/components/Messages/index.tsx
rileylnapier: its basically "useGetMessages" ~ i need to tweak this pattern and possibly rename

packages/react-provider/src/reducer.ts
rileylnapier: yeah it will.  i haven't really made or useed the toastreducer yet...

packages/react-provider/src/transports/base.ts
rileylnapier: whoops fixed.

RoyalSix (approved):

PR #7, opened 2021-03-18
[C-2468] Click Tracking Read Event

RoyalSix:
## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-provider/src/hooks/use-actions.ts
+++ b/packages/react-provider/src/hooks/use-actions.ts
@@ -1,8 +1,14 @@
+import { sendClickedRequest } from '~/lib/events';
 import useCourier from './use-courier';
 
 export default () => {
-  const { dispatch } = useCourier();
+  const { dispatch, clientKey } = useCourier();
   return {
+    handleOnClick: (data) => {
+      if (data?.clickedUrl) {
+        sendClickedRequest(clientKey, data?.clickedUrl);
+      }
+    },
     initToast: (payload) => {
       dispatch({
         type: "INIT_TOAST",
--- a/packages/react-provider/src/hooks/use-actions.ts
+++ b/packages/react-provider/src/hooks/use-actions.ts
@@ -15,7 +15,6 @@ export default () => {
         payload
       });
     },
-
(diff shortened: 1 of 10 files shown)


packages/react-toast/src/components/Body/index.tsx
```diff
@@ -23,17 +23,17 @@ const ToastBody: React.FunctionComponent<Partial<ICourierToastMessage>> = ({
 
   const handleOnClickDetails = useCallback(
     (event) => {
+      if (data?.clickedUrl) {
+        sendClickedRequest(clientKey, data?.clickedUrl);
```
rileylnapier, line 27: a clicked should cause a read instead of sending both

rileylnapier:
make sure to run `yarn lerna:version` if you plan on releasing it.  either in this PR or a separate one

packages/storybook/stories/react-toast/with-click-action.stories.js
```diff
@@ -26,8 +26,9 @@ function DefaultComponentWithAction({ body }) {
     title: "Courier",
     body,
     data: {
-      clickAction: "https://app.courier.com",
-      clickedUrl: "https://api.courier.com/something/send",
+      // clickAction: "https://app.courier.com",
```
rileylnapier, line 29: why is this commented out?

packages/react-provider/src/hooks/use-actions.ts
```diff
@@ -1,15 +1,20 @@
+import { sendClickedRequest } from '~/lib/events';
 import useCourier from './use-courier';
 
 export default () => {
-  const { dispatch } = useCourier();
+  const { dispatch, clientKey } = useCourier();
```
rileylnapier, line 5: ill make a ticket to make the toast reducer this week too since this is ambiguous what handleOnClick is

RoyalSix: Yes, I was thinking about what actions can go here that are agnostic to toast and inbox

rileylnapier (approved):

PR #9, opened 2021-03-19
Riley/c 2465 use preact for building react toast

rileylnapier:
## Description

webpack bundling of our components.  still do to deploying.

## Type of change

## Related issues

> `Fix [#1]()`


packages/embeddable/src/index.html
```diff
@@ -0,0 +1,33 @@
...
+      }
+
+      body,
+      body * {
+        box-sizing: border-box;
+      }
+    </style>
+    <script type="text/javascript">
+      window.courierConfig = {
+        clientKey: "ZDRiMTViNmMtODdmZi00OWMzLWI5YTgtYmU5NTc3YTRlM2Y1",
```
RoyalSix, line 21: Should we initialize this without userId and clientKey?

packages/embeddable/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    new HtmlWebpackPlugin({
+      template: "src/index.html",
+    }),
+    process.env.ANALYZE && new BundleAnalyzerPlugin(),
+  ].filter(Boolean),
+  module: {
+    rules: [
+      {
+        test: /\.tsx?$/,
+        exclude: /(node_modules|bower_components)/,
```
RoyalSix, line 24: Do we also need to exclude tests folder?

packages/embeddable/src/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+              return Inbox;
+            }
+          }
+        })();
+
+        if (!Component) {
+          return null;
+        }
+
+        return ReactDOM.createPortal(
```
RoyalSix, line 43: Is it common to have a bunch of different portals

packages/embeddable/src/index.html
```diff
@@ -0,0 +1,33 @@
...
+        userId: "3b17362c-0c91-4fcd-a829-885a2343eb67",
+        apiUrl: "https://rubmz24skk.execute-api.us-east-1.amazonaws.com/dev",
+        wsUrl: "wss://20en15n3ng.execute-api.us-east-1.amazonaws.com/dev",
+      };
+    </script>
+  </head>
+  <body>
+    <h1>Treva Homepage</h1>
```
RoyalSix, line 29: Think this was left over

packages/embeddable/src/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          </Suspense>,
+          element
+        );
+      })}
+    </>
+  );
+};
+
+(async () => {
+  const { clientKey, apiUrl, userId, userSignature } = window.courierConfig;
```
RoyalSix, line 55: Should we throw an error if correct props are not found on window?

RoyalSix (approved):
Will also need a new README at some point

packages/embeddable/src/index.tsx
rileylnapier: why not?  also... i don't think we have much of a choice because we need all the components to be children of the "Provider" but also exist anywhere in the dom

packages/embeddable/webpack.config.js
rileylnapier: probably

packages/embeddable/src/index.html
rileylnapier: you mean, should we support initializing async as well as sync?

PR #13, opened 2021-03-22
replace apollo with urql to slim bundle

rileylnapier:
## Description

- replace apollo-client
- fix tests
- remove better-spinkit because of lodash

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -25,7 +25,7 @@ const Messages: React.FunctionComponent<InboxProps> = ({
       )}
       <Body data-testid="messages">
         {isLoading ? (
-          <Loading size={100} color="#9E3789" />
+          <Loading />
```
rileylnapier, line 28: need a new loading after removing betterspinkit

cgradwohl (approved):

PR #15, opened 2021-03-22
lazy tippy and loading skeleton

rileylnapier:
## Description

image

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/components/Inbox/LazyTippy.tsx
+++ b/packages/react-inbox/src/components/Inbox/LazyTippy.tsx
@@ -8,7 +8,6 @@ const LazyTippy: React.FunctionComponent<TippyProps> = forwardRef(
     const lazyPlugin = {
       fn: () => ({
         onMount: () => setMounted(true),
-        onHidden: () => setMounted(false),
       }),
     };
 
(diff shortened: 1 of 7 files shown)


packages/react-inbox/src/components/Inbox/LazyTippy.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const lazyPlugin = {
+      fn: () => ({
+        onMount: () => setMounted(true),
+        onHidden: () => setMounted(false),
+      }),
+    };
+
+    const computedProps = { ...props };
+
+    computedProps.plugins = [lazyPlugin, ...(props.plugins || [])];
```
RoyalSix, line 17: `lazyPlugin` is an object, so computedProps.plugins is an array of objects.
But I am confused on what `...(props.plugins || [])` resolves to.
 Maybe some typedefs here

packages/react-inbox/src/components/Inbox/LazyTippy.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { forwardRef, useState } from "react";
+import Tippy, { TippyProps } from "@tippyjs/react";
+
+const LazyTippy: React.FunctionComponent<TippyProps> = forwardRef(
+  (props, ref) => {
+    const [mounted, setMounted] = useState(false);
+
+    const lazyPlugin = {
+      fn: () => ({
```
RoyalSix, line 9: Could we add a comment here what `fn` means? or how it will be used

rileylnapier: ill add a link to the gist i stole it from, i think its part of how you write tippy plugins

RoyalSix (approved):

PR #24, opened 2021-03-26
trackEvent mutation and tabs

rileylnapier:
## Description

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/components/OptionsDropdown/Options.tsx
+++ b/packages/react-inbox/src/components/OptionsDropdown/Options.tsx
@@ -31,14 +31,22 @@ const Option = styled.button`
   }
 `;
 
-function Options({ options }) {
+function Options({ options, onClose }) {
   return (
     <Container>
-      {options.map(({ label, onClick }) => (
-        <Option key={label} onClick={onClick}>
-          {label}
-        </Option>
-      ))}
+      {options.map(({ label, onClick }) => {
+        const handleOnClick = (event: React.MouseEvent) => {
+          event.preventDefault();
+          onClick();
+          onClose();
+        };
+
+        return (
+          <Option key={label} onClick={handleOnClick}>
+            {label}
+          </Option>
+        );
+      })}
     </Container>
   );
 }
--- a/packages/react-inbox/src/components/OptionsDropdown/Options.tsx
+++ b/packages/react-inbox/src/components/OptionsDropdown/Options.tsx
@@ -34,7 +34,7 @@ const Option = styled.button`
 function Options({ options }) {
   return (
     <Container>
-      {options.map(({ label, onClick = () => {} }) => (
+      {options.map(({ label, onClick }) => (
         <Option key={label} onClick={onClick}>
           {label}
         </Option>
--- a/packages/react-inbox/src/components/OptionsDropdown/Options.tsx
+++ b/packages/react-inbox/src/components/OptionsDropdown/Options.tsx
@@ -1,5 +1,5 @@
-import React from 'react'
-import styled from 'styled-components';
+import React from "react";
+import styled from "styled-components";
 
 const Container = styled.div`
   display: flex;
@@ -9,7 +9,7 @@ const Container = styled.div`
   border-radius: 4px !important;
 `;
 
-const Option = styled.div`
+const Option = styled.button`
   height: 25px;
   font-size: 12px;
   font-style: normal;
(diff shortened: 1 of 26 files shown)


packages/react-inbox/src/components/OptionsDropdown/Options.tsx
```diff
@@ -22,23 +22,25 @@ const Option = styled.div`
...
   return (
     <Container>
-      {
-        options.map(({ label, onClick = () => {} }) => <Option key={label} onClick={() => {
-          onClick();
-          close()
-        }}>{label}</Option>)
-      }
+      {options.map(({ label, onClick = () => {} }) => (
+        <Option key={label} onClick={onClick}>
```
rileylnapier, line 38: don't onclick a div

packages/react-inbox/src/components/OptionsDropdown/index.tsx
```diff
@@ -29,24 +29,22 @@ const StyledTippy = styled(Tippy)`
...
-  offset: [0, -8]
+  offset: [0, -8],
 };
 
 function OptionsDropdown({ options }) {
-  const [showDropdown, setShowDropdown] = useState(false);
-  
   return (
     <Container>
-      <StyledTippy {...tippyProps} visible={showDropdown} content={<Options close={() => setShowDropdown(false)} options={options} />}>
```
rileylnapier, line 43: tippy was getting mad by having "visible" and "trigger" props at the same time

packages/react-inbox/src/hooks/use-inbox.ts
```diff
@@ -1,47 +1,144 @@
...
+    transport?.listen({
+      id: "inbox-listener",
+      listener: (courierEvent) => {
+        newMessage(courierEvent?.data);
+      },
+    });
+  }, [transport]);
 
   return {
     ...inbox,
```
RoyalSix, line 37: Would like to see how we can get a more declarative pattern for this, I keep getting confused about whats actually in the `inbox` object. (Don't need to solve this here)

packages/react-inbox/src/hooks/use-inbox.ts
```diff
@@ -1,47 +1,144 @@
...
-        type: "inbox/NEW_MESSAGE",
-        payload,
+        type: "inbox/SET_ALL_MESSAGES",
+        payload
+      });
+    },
+
+    markMessageRead: async (messageId: string, trackingId: string) => {
+      await trackEvent({
+        trackingId
```
RoyalSix, line 118: Nice abstraction 👍🏾

packages/react-inbox/src/hooks/use-inbox.ts
```diff
@@ -1,47 +1,144 @@
...
       });
     },
+
+    markMessageUnread: async (messageId: string, trackingId: string) => {
+      await trackEvent({
+        trackingId
+      });
+
+      dispatch({
+        type: "inbox/MARK_MESSAGE_UNREAD",
```
RoyalSix, line 135: Would like to have this as a constant namespace in the future

RoyalSix (dismissed):

packages/react-inbox/src/hooks/use-inbox.ts
rileylnapier: i feel the same, right now its basically the main props passed into the component

packages/react-inbox/src/hooks/use-inbox.ts
rileylnapier: agree

packages/react-provider/src/hooks/use-graphql-client.ts
```diff
@@ -8,6 +8,7 @@ export default () => {
   const client = useMemo(() => {
     return createClient({
       url: `${apiUrl ?? process.env.API_URL ?? `https://api.courier.com`}/client/q`,
+      requestPolicy: "network-only",
```
rileylnapier, line 11: this will need to change

packages/react-inbox/src/reducer.ts
```diff
@@ -70,10 +70,13 @@ export default (state: InboxState = {}, action) => {
   }
 
   case "inbox/MARK_MESSAGE_READ": {
+    const unreadMessageCount = (state.unreadMessageCount ?? 1) - 1;
```
RoyalSix, line 73: Not sure I understand why its -1 but approving. Maybe we could get a `Math.max(unreadMessageCount, 0)`

RoyalSix (approved):

packages/react-inbox/src/reducer.ts
rileylnapier: This is a good idea

PR #28, opened 2021-03-30
[C-2560] Mark As Unread

RoyalSix:
## Description
As you can see below the unread functionality is now connected the front end mutation.
There is a separate issue regarding caching and reloading that also needs to be fixed

https://www.loom.com/share/c1371b1d508f44878ad29ce3e4ad5316

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -58,44 +60,44 @@ const Message: React.FunctionComponent<MessageProps> = ({
...
     () =>
       [
         !read &&
           trackingIds?.readTrackingId && {
-            label: "Mark as Read",
-            onClick: () => {
-              if (!trackingIds?.readTrackingId) {
-                return;
-              }
+          label: "Mark as Read",
```
rileylnapier, line 73: this spacing looks off

packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -58,44 +60,44 @@ const Message: React.FunctionComponent<MessageProps> = ({
...
+            markMessageUnread(messageId, trackingIds?.unreadTrackingId);
           },
+        },
         /*{
         label: "Delete",
         onClick: () => {},
       },*/
       ].filter(Boolean),
-    [read, messageId, trackingIds]
+    [read, trackingIds?.readTrackingId, trackingIds?.unreadTrackingId, markMessageRead, messageId, markMessageUnread],
```
rileylnapier, line 99: we need to add `markMessageUnread` and `markMessageRead` and each tracking id separatly here?

packages/react-inbox/src/components/Message/index.tsx
RoyalSix: Screen Shot 2021-03-30 at 2 24 59 PM

My editor gives me an error. Wonder if we have different settings

rileylnapier: we should get consistent here, newline with an object should be indented

rileylnapier (dismissed):

packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -58,44 +61,35 @@ const Message: React.FunctionComponent<MessageProps> = ({
...
-            label: "Mark as Read",
-            onClick: () => {
-              if (!trackingIds?.readTrackingId) {
-                return;
-              }
-
-              markMessageRead(messageId, trackingIds?.readTrackingId);
-            },
+        showMarkAsRead && {
+          label: "Mark as Read",
```
suhasdeshpande, line 74: are we sure okay to be hard coding `Mark as Read` string?

packages/react-inbox/src/hooks/use-inbox.ts
```diff
@@ -36,33 +38,33 @@ export default () => {
...
+          }],
+        },
       });
     },
 
     setCurrentTab: (newTab: string) => {
       dispatch({
         type: "inbox/SET_CURRENT_TAB",
-        payload: newTab
+        payload: newTab,
```
suhasdeshpande, line 55: lot of lint changes, did we change some rules recently? if so, can we break this PR into lint changes and code change relevant to marking as unread?

packages/react-inbox/src/reducer.ts
```diff
@@ -64,8 +64,27 @@ export default (state: InboxState = {}, action) => {
...
         ...(state.messages ??[]),
-        ...newMessages
-      ] : newMessages
+        ...newMessages,
+      ] : newMessages,
+    };
+  }
+
+  case "inbox/MARK_MESSAGE_UNREAD": {
+    const unreadMessageCount = (state.unreadMessageCount ?? 1) + 1;
```
suhasdeshpande, line 73: could we make this sort of decision on the dispatch end? Like could this be part of action payload rather than reducer logic?

packages/react-inbox/src/reducer.ts
```diff
@@ -64,8 +64,27 @@ export default (state: InboxState = {}, action) => {
...
+
+    return {
+      ...state,
+      messages: state.messages?.map(message => {
```
suhasdeshpande, line 77: why are we not sure that messages are legit array? When we dispatch an action, we would have sensible default / initial state right?

packages/react-provider/src/index.tsx
```diff
@@ -19,7 +19,7 @@ export const registerReducer = _registerReducer;
 export type ICourierMessage = TransportTypes.ICourierMessage;
 export type ICourierContext = types.ICourierContext;
 export const CourierContext = React.createContext<ICourierContext | undefined>(
```
suhasdeshpande, line 21: could you tell me how setting context to `undefined` different than not setting anything?

packages/react-toast/src/components/Body/index.tsx
```diff
@@ -19,14 +21,15 @@ const ToastBody: React.FunctionComponent<Partial<ICourierToastMessage>> = ({
...
   );
   const handleOnClickDetails = useCallback((event) => {
-    handleOnClick(data)
+    handleOnClick(data);
+
     if (onClick) {
       onClick(event);
     }
-  }, [])
+  }, [data, handleOnClick, onClick]);
```
suhasdeshpande, line 32: why would a pointer to a function like handleOnClick or onClick change?

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -101,8 +106,9 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
           </span>
         ) : (
           <Bell
-            hasUnreadMessages={Boolean(inbox.unreadMessageCount)}
             className={props.className}
+            hasUnreadMessages={Boolean(inbox.unreadMessageCount)}
+            onMouseEnter={handleBellOnMouseEnter}
```
rileylnapier, line 111: i don't think we want this

rileylnapier: we only have one current list of messages, so until we have separate lists we can't prefetch

packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -71,22 +72,18 @@ const Message: React.FunctionComponent<MessageProps> = ({
     () =>
       [
         showMarkAsRead && {
-          label: "Mark as Read",
+          label: MESSAGE_LABELS.MARK_AS_READ,
```
rileylnapier, line 75: if these labels are only used in one place, does making constants file do anything for us or is it just something "to do"?

rileylnapier: my personal persepctive is that if something is used >1 or maybe event >2 times then abstract it out... but if its only a one time constant or a a one time peice of logic than writing a separate file/function just adds unnecesary complexity

rileylnapier: my opinion though, so feel free to disregard

packages/react-toast/src/components/Actions/index.tsx
```diff
@@ -19,7 +21,7 @@ const Actions: React.FunctionComponent<{
       {onClickDetails && (
         <Button
           href={href}
-          color="#9D3789"
+          color="#9121c2"
```
rileylnapier, line 24: what color change is this?

packages/react-toast/src/components/Toast/styled.ts
```diff
@@ -4,7 +4,9 @@ import { ToastContainer } from "react-toastify";
 export const ToastStyled = styled(ToastContainer)(({ theme }) => ({
   ["&.Toastify__toast-container"]: {
     fontFamily: `"Nunito", sans-serif`,
-    ...theme?.root
+    marginTop:150,
```
rileylnapier, line 7: spacing

packages/react-toast/src/components/Toast/styled.ts
```diff
@@ -19,7 +21,7 @@ export const ToastStyled = styled(ToastContainer)(({ theme }) => ({
     ...theme?.body
   },
   [".Toastify__progress-bar"]: {
-    background: "rgb(157, 55, 137)",
+    background: "#9121c2",
```
rileylnapier, line 24: can we screenshot the updates to style?

packages/react-inbox/src/components/Message/index.tsx
rileylnapier: cool cool, just sharing my thoughts.  we should add this to the list of things we want to do as a part of our contributing guide we will someday make.  me you and @suhasdeshpande  should probably have a meeting to come up with some best practices

packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -43,7 +43,7 @@ const Messages: React.FunctionComponent<InboxProps> = ({
   useEffect(() => {
     fetchMessages(currentTab?.filter);
   }, [currentTab]);
-
+  console.log("messages", messages);
```
rileylnapier, line 46: remove

packages/react-inbox/src/reducer.ts
rileylnapier: this should be `const unreadMessageCount = (state.unreadMessageCount ?? 0) + 1;`

the other way was to prevent it going to `-1`

rileylnapier: but i liked your idea on Max/Min

packages/react-provider/src/index.tsx
rileylnapier: typescript got mad if you didn't.  basically you have 2 choices.  make the context with a context object or make it with undefined.  There isn't a way to say ICourierContext is optional `React.createContext<ICourierContext?>` so you have to `| undefined`

RoyalSix: Makes sense. Resolving

packages/react-inbox/src/reducer.ts
RoyalSix: Fixed.

packages/react-inbox/src/reducer.ts
RoyalSix: Nice catch. Thank you. Fixed

packages/react-inbox/src/hooks/use-inbox.ts
RoyalSix: Removed.

packages/react-inbox/src/components/Message/index.tsx
RoyalSix: Fixed.

packages/react-toast/src/components/Toast/styled.ts
RoyalSix: Removed.

packages/react-inbox/src/components/Messages/index.tsx
RoyalSix: Done.

rileylnapier (approved):

PR #32, opened 2021-04-05
localstorage

rileylnapier:
## Description

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/components/Inbox/index.tsx
+++ b/packages/react-inbox/src/components/Inbox/index.tsx
@@ -82,7 +82,30 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
   }, []);
 
   useEffect(() => {
-    inbox.init(props);
+    const { clientKey, userId, inbox } = courierContext;
+
+    if (!inbox && clientKey && userId) {
+      const localStorageState = localStorage.getItem(
+        `${clientKey}/${userId}/inbox`
+      );
+
+      if (localStorageState) {
+        try {
+          inbox.init({
+            ...JSON.parse(localStorageState),
+            config: props,
+          });
+        } catch {
+          // do nothing
+        }
+      }
+
+      return;
+    }
+
+    inbox.init({
+      config: props,
+    });
   }, [props]);
 
   const handleBellOnMouseEnter = (event: React.MouseEvent) => {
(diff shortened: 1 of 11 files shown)


packages/react-toast/src/components/Toast/index.tsx
```diff
@@ -23,39 +24,38 @@ export const Toast: React.FunctionComponent<{
...
 
   const handleToast = useCallback(
     (message: ICourierToastMessage | string) => {
       message =
         typeof message === "string"
           ? ({
-            body: message,
-            icon: undefined,
-          } as ICourierToastMessage)
+              body: message,
```
suhasdeshpande, line 34: extra spacing

packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -67,16 +71,21 @@ const Messages: React.FunctionComponent<InboxProps> = ({
...
             <Message key={message.messageId} {...message} />
-          ),
+          )
         )}
         {isLoading && <Loading />}
         {!isLoading && messages.length === 0 && (
           <Empty>You have no notifications at this time</Empty>
         )}
-        {!isLoading && messages.length && !startCursor && (<PaginationEnd title="End Of The Road" />) }
+        {!isLoading && messages.length > 5 && !startCursor && (
```
rileylnapier, line 80: im not too sure about this logic here, but i think we only want to show the end of the road if we have > 5 messages

packages/react-toast/src/components/Toast/index.tsx
rileylnapier: we need to get the linter right on this stuff.

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -82,8 +84,46 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
+          console.log("error", ex);
+          // do nothing
+        }
+      }
+    }
+
+    initInbox({
+      config: props,
+    });
+  }, [props, clientKey, userId]);
```
RoyalSix, line 109: Should this be getting called everytime the props object changes?

rileylnapier: yeah, because we need to update the context

RoyalSix (approved):

PR #34, opened 2021-04-06
Version17

rileylnapier:
## Description

- fix unread count
- fix setting current tab on load

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/hooks/use-inbox.ts
+++ b/packages/react-inbox/src/hooks/use-inbox.ts
@@ -4,18 +4,19 @@ import useMessages from "~/hooks/use-messages";
 
 export default () => {
   const { fetch: fetchMessages } = useMessages();
-  const {
-    dispatch, inbox, transport,
-  } = useCourier();
+  const { dispatch, inbox, transport } = useCourier();
 
   const [_, trackEvent] = useTrackEvent();
 
-  const newMessage = useCallback((payload) => {
-    dispatch({
-      type: "inbox/NEW_MESSAGE",
-      payload,
-    });
-  }, [dispatch]);
+  const newMessage = useCallback(
+    (payload) => {
+      dispatch({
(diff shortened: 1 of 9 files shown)


packages/react-inbox/src/reducer.ts
```diff
@@ -29,13 +29,13 @@ const initialState: InboxState = {
...
 export default (state: InboxState = initialState, action) => {
   switch (action.type) {
-  case "inbox/INIT": {
-    return {
-      ...state,
-      ...action.payload,
-      currentTab: action.payload?.tabs?.[0],
-    };
-  }
+    case "inbox/INIT": {
```
suhasdeshpande, line 32: something seems to be happening here with extra spaces

packages/react-inbox/src/hooks/use-inbox.ts
```diff
@@ -4,18 +4,19 @@ import useMessages from "~/hooks/use-messages";
...
-    });
-  }, [dispatch]);
+  const newMessage = useCallback(
+    (payload) => {
+      dispatch({
+        type: "inbox/NEW_MESSAGE",
+        payload,
+      });
+    },
+    [dispatch]
```
suhasdeshpande, line 18: do we need useCallback here?

rileylnapier: because everytime we called this function, it was creating a new function and causing extra rerenders. 

also, eslint told me to

suhasdeshpande (approved):
seems fine. some odd spacing and one question about useCallback.

packages/react-inbox/src/reducer.ts
rileylnapier: when i rebased the lint changes into my previous branch, i didn't re-run the prettier on the files.  so this changes now applied them

PR #35, opened 2021-04-07
[C-2565] Mark All As Read

RoyalSix:
## Description

https://user-images.githubusercontent.com/16184018/113905747-54dc4400-9788-11eb-84d2-e2f848e1e0c3.mp4

Making a separate issue to fix some more caching/usability issues

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -64,19 +64,20 @@ const Message: React.FunctionComponent<MessageProps> = ({
...
-        label: "View Details",
-        onClick: () => {
-          if (trackingIds?.clickTrackingId) {
-            trackEvent({
-              trackingId: trackingIds?.clickTrackingId,
-            });
-          }
+    () =>
+      [
+        data?.clickAction && {
```
RoyalSix, line 69: Conditionally rendering this button. If there is no `clickAction` the button does nothing

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
rileylnapier, line 1: cool

packages/react-provider/src/hooks/use-track-event.ts
```diff
@@ -8,8 +8,18 @@ const TRACK_EVENT = `
...
+      ids
+    }
+  }
+`;
+
 const useTrackEvent = () => {
-  return useMutation(TRACK_EVENT);
+  const [, trackEvent] = useMutation(TRACK_EVENT);
+  const [, batchTrackEvent] = useMutation(BATCH_TRACK_EVENT);
+  return [trackEvent, batchTrackEvent];
```
rileylnapier, line 22: i feel like we should just return an object here

rileylnapier (approved):

packages/react-provider/src/hooks/use-track-event.ts
RoyalSix: Agree

rileylnapier (approved):

PR #37, opened 2021-04-08
tweak the side-effects

suhasdeshpande:
## Description

I am not certain if returning from useEffect is causing the component to not handle the side-effects correctly. 

## Type of change

### DIFF
--- a/packages/react-provider/src/index.tsx
+++ b/packages/react-provider/src/index.tsx
@@ -74,7 +74,7 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
   }, [apiUrl, clientKey, transport, userId, userSignature]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/react-provider/src/index.tsx
```diff
@@ -74,18 +74,17 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
   }, [apiUrl, clientKey, transport, userId, userSignature]);
 
   useEffect(() => {
-    if (!transport || !userId) {
-      return;
+    if(transport && userId) {
```
suhasdeshpande, line 77:
```suggestion
    if (transport && userId) {
```

RoyalSix:
The test is failing due to a type issue. 
should be `(transport as CourierTransport)`

PR #38, opened 2021-04-12
Add Brand ID To Provider Config

RoyalSix:
## Description

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/constants.ts
```diff
@@ -2,3 +2,18 @@ export const MESSAGE_LABELS = {
   MARK_AS_READ: "Mark as Read",
   MARK_AS_UNREAD: "Mark as Unread",
 };
+
+export const DEFAULT_TABS = [
```
rileylnapier, line 6: 👍

packages/react-provider/src/hooks/use-brand.ts
```diff
@@ -0,0 +1,32 @@
...
+    const [results] = useQuery({
+      query: GET_BRAND,
+      variables: {
+        brandId,
+      },
+    });
+    debugger;
+    setBrand(results?.data);
+  }
+  console.log("brand", brand);
```
rileylnapier, line 28: remove?

packages/storybook/stories/react-inbox/with-toast.stories.tsx
```diff
@@ -52,6 +52,7 @@ export function Default() {
         clientKey={clientKey}
         userId={userId}
         transport={courierTransport}
+        brandId="S0E8PFMXC44SSFG6J8A7EF4QXRNX"
```
rileylnapier, line 55: might want an env variable so we can change between our dev envs

RoyalSix: Done

packages/react-provider/src/hooks/use-brand.ts
RoyalSix: Done

packages/react-provider/src/index.tsx
```diff
@@ -52,29 +58,37 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
...
+          apiUrl,
+          brandId,
+          clientKey,
+          transport,
+          userId,
+          userSignature,
+        },
+      });
+      if (brandId) {
+        dispatch({
```
rileylnapier, line 85: this should be in its own useEffect

packages/react-provider/src/reducer.ts
```diff
@@ -33,6 +33,13 @@ const rootReducer = (state, action) => {
         transport: action.payload.transport,
         userId: action.payload.userId,
         userSignature: action.payload.userSignature,
+        brandId: action.payload.brandId,
+      };
+    }
+    case "root/GET_BRAND/DONE": {
```
rileylnapier, line 39: what do we do while we are fetching a brand?  we prob don't want to render anything right?

packages/react-provider/src/index.tsx
```diff
@@ -88,15 +102,14 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
...
-
   return (
     <CourierContext.Provider
       value={{
         ...(state as any),
         dispatch,
       }}
     >
-      <GraphQLProvider>{children}</GraphQLProvider>
+      <GraphQLProvider client={client}>{children}</GraphQLProvider>
```
rileylnapier, line 112: why do we need to pass client in here?

rileylnapier (approved):

PR #42, opened 2021-04-15
Jay/c 2702 definedocument state management for

RoyalSix:
## Description

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/actions/message-count.ts
```diff
@@ -0,0 +1,12 @@
+export const QUERY_MESSAGE_COUNT = `
+  query MessageCount($isRead: Boolean!) {
+    messageCount(params: {
+      isRead: $isRead
+    })
+  }
+`;
+
+export const getMessageCount = async (client, { isRead = false }) => {
```
rileylnapier, line 9: do we have to pass client in?

packages/react-provider/src/graph-ql/index.ts
```diff
@@ -1,9 +1,9 @@
-import { useMemo } from "react";
-import { createClient } from "urql";
+import { Client, createClient } from "urql";
 
-export default ({ clientKey, userId, userSignature, apiUrl }) => {
-  const client = useMemo(() => {
-    return createClient({
+export class GraphQLClient {
+  client: Client;
+  constructor({ clientKey, userId, userSignature, apiUrl }) {
```
rileylnapier, line 5: what happens if the userId changes?

packages/react-provider/src/middleware.ts
```diff
@@ -9,7 +9,7 @@ const asyncMiddleware = (store) => (next) => async (action) => {
   });
 
   try {
-    const result = await action.payload();
+    const result = await action.payload(store.getState());
```
rileylnapier, line 12: this feels a little weird

packages/storybook/stories/react-inbox/with-toast.stories.tsx
```diff
@@ -49,6 +50,8 @@ export function Default() {
   return (
     <>
       <CourierProvider
+        apiUrl={API_URL}
+        // wsUrl={WS_URL}
```
rileylnapier, line 54: do we need wsurl?

packages/react-inbox/src/actions/message-count.ts
RoyalSix: I believe so, I cannot think of another pattern without using hooks, or being on the window object

packages/react-provider/src/graph-ql/index.ts
RoyalSix: If any one these variables change a new client will be created. Check here https://github.com/trycourier/courier-react/blob/d66fbf0e9f59eb5fdc7b8341487fa20729a50863/packages/react-provider/src/index.tsx#L36

packages/react-provider/src/middleware.ts
RoyalSix: I updated to use `react-thunk` pattern

packages/storybook/stories/react-inbox/with-toast.stories.tsx
RoyalSix: In the example, no. I removed

rileylnapier (approved):
i think we're getting closer.  still a few things that feel off but good to go

PR #49, opened 2021-04-20
branding

rileylnapier:
## Description

- get default brand
- apply branding for 1/2/3 colors
- save brand in local storage

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Inbox/tippy.css
```diff
@@ -0,0 +1,72 @@
+.tippy-box[data-animation="fade"][data-state="hidden"] {
```
rileylnapier, line 1: forked css so we can control what default tippy css is

packages/react-inbox/src/components/Actions/styled.ts
```diff
@@ -11,21 +11,28 @@ export const Action = styled.a<{ href: string; target: string }>(({ theme }) =>
       cursor: "pointer",
       border: "none",
       fontSize: 12,
-      color: theme?.brand?.primaryColor ?? "#9121C2",
+      color: theme?.brand?.inapp?.colors?.invertButtons
```
suhasdeshpande, line 14: do we need to default to anything? is ther any point after which we are 100% certain of things existing, to avoid `?`?

rileylnapier: we will for all legacy brands that don't have any colors set

scarney81 (approved):

packages/react-inbox/src/components/Actions/styled.ts
rileylnapier: added this: https://github.com/trycourier/backend/pull/1500

PR #51, opened 2021-04-20
fix subscription

rileylnapier:
## Description

this.connected was always undefined

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-provider/src/ws.ts
+++ b/packages/react-provider/src/ws.ts
@@ -5,6 +5,7 @@ export class WS {
   protected connected;
   protected messageCallback;
   private url: string;
+
   constructor({ url }) {
     this.messageCallback = null;
     this.connection = undefined;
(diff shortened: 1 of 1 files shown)


packages/react-provider/src/ws.ts
```diff
@@ -14,29 +15,25 @@ export class WS {
 
   connect(clientKey: string): void {
     const url = `${this.url}/?clientKey=${clientKey}`;
-
-    if (typeof WebSocket) {
-      this.connection = new WebSocket(url);
-      this.initiateListener();
-    }
+    this.connection = new WebSocket(url);
```
suhasdeshpande, line 18: do we not need to check for Websocket? guessing if the react is run on server-side, it will fail, no?

rileylnapier: id rather have it fail than silently do nothing

suhasdeshpande (approved):

PR #63, opened 2021-04-27
[C-2851] adds onmessage prop to CourierProvider

rileylnapier:
## Description

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -10,77 +10,79 @@ import { useAtBottom } from "~/hooks/use-at-bottom";
...
-    fetchMessages,
-    markAllAsRead,
-    currentTab,
-    isLoading,
-    messages = [],
-    startCursor,
-    unreadMessageCount,
-  } = useInbox();
-  const ref = useRef<HTMLDivElement>(null);
+const Messages: React.ForwardRefExoticComponent<
```
suhasdeshpande, line 13: what makes this exotic? what are the advantages I mean? I read up on it on StackOverflow but I wasn't clear if I understand the use-case

> Also, a big difference between the two definitions seems to be, that ForwardRefExoticComponent (like all exotic components) are no function components, but actually just objects, which are treated specially when rendering them. Therefore the comment

packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -10,77 +10,79 @@ import { useAtBottom } from "~/hooks/use-at-bottom";
...
-      </MessageList>
-      <Footer>
-        Powered by&nbsp;&nbsp;
-        <CourierLogo />
-      </Footer>
-    </>
-  );
-};
+        <TabList />
+        <MessageList ref={divRef} data-testid="messages">
```
suhasdeshpande, line 63: could we use more explicit name than `divRef`?

packages/react-inbox/src/hooks/use-click-outside.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from "react";
+
+export default (ref, handler) => {
```
suhasdeshpande, line 3: could we use named export? if not, I would still suggest adding named functions, and then default export that. It helps with future debugging.

packages/react-provider/src/index.tsx
```diff
@@ -63,6 +66,28 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
     middleware,
   });
 
+  useEffect(() => {
+    if (_transport) {
+      // this means the transport was passed in and we shouldn't subscribe
```
suhasdeshpande, line 71: what are the scenarios where one would pass in transport?

packages/react-inbox/src/hooks/index.ts
```diff
@@ -1 +1,2 @@
 export { default as useInbox } from "./use-inbox";
+export { default as useClickOutside } from "./use-click-outside";
```
suhasdeshpande, line 2: why not just named export `useClickOutside` instead of aliasing default export?

packages/storybook/stories/inbox/with-toast.stories.tsx
```diff
@@ -99,18 +104,22 @@ export function AsyncUserId() {
   );
 }
 
-export function WithTheme() {
+export function OnMessage() {
+  const handleOnMessage = (message) => {
+    console.log(message);
```
suhasdeshpande, line 109: probably remove log?

packages/react-provider/src/types.ts
```diff
@@ -32,4 +33,5 @@ export interface ICourierContext {
   userSignature?: string;
   wsUrl?: string;
   middleware?: any;
+  onMessage?: Interceptor;
```
suhasdeshpande, line 36: `onMessage` interface requires a return value, is that useful or can it be `void`? we don't seem to be using the return value from onMessage?

packages/react-provider/src/index.tsx
```diff
@@ -40,16 +41,18 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
     return new Client({ clientKey, userId, userSignature, apiUrl });
   }, [clientKey, userId, userSignature, apiUrl]);
 
-  const courierTransport = useMemo(() => {
+  const transport = useMemo(() => {
```
suhasdeshpande, line 44: can we do something like this instead?  
```suggestion
  const transport = useMemo< CourierTransport >(() => {
```

packages/react-provider/src/index.tsx
```diff
@@ -63,6 +66,28 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
...
+    if (_transport) {
+      // this means the transport was passed in and we shouldn't subscribe
+      return;
+    }
+
+    if (!transport || !userId) {
+      return;
+    }
+
+    const courierTransport = transport as CourierTransport;
```
suhasdeshpande, line 79: https://github.com/trycourier/courier-react/pull/63/files#r621480088

packages/react-inbox/src/hooks/use-click-outside.ts
suhasdeshpande, line 3:
```suggestion
export default (ref, handler): void => {
```

packages/react-inbox/src/components/Messages/index.tsx
rileylnapier: its just the typescript name for it.

packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -10,77 +10,79 @@ import { useAtBottom } from "~/hooks/use-at-bottom";
...
+    } = useInbox();
+    const divRef = useRef<HTMLDivElement>(null);
 
-  useAtBottom(
-    ref,
-    () => {
-      if (isLoading || !startCursor) {
-        return;
-      }
+    useAtBottom(
```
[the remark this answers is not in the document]
rileylnapier: because of the React.forwardRef

packages/react-inbox/src/hooks/index.ts
rileylnapier: idk, no reason

packages/react-provider/src/index.tsx
rileylnapier: that was originally how it all worked, you had to create and pass in the transport.  so we have many storybook examples doing it this way.  im not sure if it will happen in the wild... but it could

packages/react-provider/src/index.tsx
rileylnapier: whats this?

packages/react-provider/src/types.ts
rileylnapier: we are, i pass onMessage in as `courier.intercept(onMessage)` and if onMessage returns without a message, it will prevent a message from showing

packages/storybook/stories/inbox/with-toast.stories.tsx
rileylnapier: eh, its in a story

rileylnapier: ill be revisting all of these stories soon

packages/react-provider/src/index.tsx
suhasdeshpande: this was a suggestion to avoid explicit casting transport to `CourierTransport`

suhasdeshpande (approved):

packages/react-provider/src/index.tsx
rileylnapier: is there a differet way?

PR #66, opened 2021-04-29
InApp Branding Update

rileylnapier:
## Description

- Ability to import ToastBody so we can display in Studio/Brand
- Allow Toast to pass in brand
- Have Toast/Inbox use remote Brand config
- Style Tweaks

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-toast/src/components/index.tsx
+++ b/packages/react-toast/src/components/index.tsx
@@ -21,6 +21,8 @@ export const ToastBody: React.FunctionComponent<ICourierToastMessage> = (
           className="Toastify__progress-bar Toastify__progress-bar--animated Toastify__progress-bar--default"
           style={{
             animation: "none",
+
+            // will want to tweak these after we get them inside studio
             //animationDuration: "5000ms",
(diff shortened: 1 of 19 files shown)


packages/react-toast/src/components/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <div className="Toastify__toast Toastify__toast--default">
+        <div className="Toastify__toast-body">
+          <Body {...props} />
+        </div>
+        <div
+          className="Toastify__progress-bar Toastify__progress-bar--animated Toastify__progress-bar--default"
+          style={{
+            animation: "none",
+
+            // will want to tweak these after we get them inside studio
```
suhasdeshpande, line 25: do we need to keep the commented block?

suhasdeshpande (approved):

packages/react-toast/src/components/index.tsx
rileylnapier: i wanna tweak these settings as i play with it in studio

PR #69, opened 2021-04-30
[C-2873]  adds preferences in inbox component

suhasdeshpande:
## Type of change


packages/react-inbox/src/components/Messages/Header/index.tsx
```diff
@@ -1,23 +1,81 @@
...
     <Container data-testid="header">
-      {title}
-      {unreadMessageCount ? ` (${unreadMessageCount})` : ""}
-      {currentTab?.filter?.isRead === false && messages.length > 0 && (
-        <MarkAllAsRead onClick={markAllAsRead} style={{ cursor: "pointer" }}>
-          Mark all as read
-        </MarkAllAsRead>
-      )}
+      {heading}
+      <SettingsIconButton onClick={() => onViewToggle("settings")}>
```
rileylnapier, line 75: can we not do anon functions.  lets name them

rileylnapier: i don't know why i think this.  i remember having aversion to anonmous functions in a render... but for onclicks im not sure exactly.

rileylnapier (approved):

packages/react-inbox/src/components/Messages/Header/index.tsx
suhasdeshpande: I can clear that up. sure I will be touching this section soon.

PR #72, opened 2021-05-05
InApp Updates

rileylnapier:
## Description

- Fix Types (typescript doesn't like `export *`
- Bell Icon styles
- Make sure to Init if clientkey and user id are missing
- Fix message icon with brand
- Match unreadindicator colors
- Fix toast brand styling

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-preferences/src/hooks/use-preferences.ts
```diff
@@ -45,7 +45,7 @@ export const usePreferenceTemplate = (
     IPreferenceTemplate | undefined
   >(undefined);
 
-  const context = useContext<ICourierContext | undefined>(CourierContext);
+  const context = useCourier();
```
rileylnapier, line 48: @suhasdeshpande fyi we have a hook for this

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -108,13 +114,16 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
+          didInit = true;
         } catch (ex) {
           console.log("error", ex);
         }
-      } else {
-        init(props);
       }
     }
+
+    if (!didInit) {
```
suhasdeshpande, line 124: do we want this boolean outside of effect callback?

packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -44,7 +44,10 @@ const Message: React.FunctionComponent<IMessageProps> = ({
       3. from props.brand.inapp.icons.message
       4. from remote brand.inapp.icons.message
     */
-    icon ?? defaultIcon ?? !brand?.inapp?.disableMessageIcon
+    icon ||
+      brand?.inapp?.icons?.message ||
```
suhasdeshpande, line 48: `?.message` for icon? maybe I am not reading this right but it is bit hard to understand.

packages/react-inbox/src/components/Message/styled.tsx
```diff
@@ -107,7 +107,7 @@ export const UnreadIndicator = styled.div(({ theme }) =>
   deepExtend(
     {
       height: "36px",
-      backgroundColor: theme?.brand?.colors?.secondary ?? "#9121c2",
+      backgroundColor: theme?.brand?.colors?.tertiary ?? "#9121c2",
```
suhasdeshpande, line 110: slightly tangent, do we have a pattern to take this hex values and put in some common constant file maybe? if we ever decide to change this hex, we will have to look for all occurrences.

packages/react-inbox/src/components/Messages/Header/index.tsx
```diff
@@ -45,7 +45,7 @@ const Header: React.FunctionComponent<IHeaderProps> = ({
                 {title}
                 {unreadMessageCount ? ` (${unreadMessageCount})` : ""}
               </h3>
-              <MessageSubHeader>Here is what you missed</MessageSubHeader>
+              {/*<MessageSubHeader>Here is what you missed</MessageSubHeader>*/}
```
suhasdeshpande, line 48: are we not ready to expose this change yet?

packages/react-toast/src/components/index.tsx
```diff
@@ -1,34 +1,58 @@
...
 const Styled = styled.div(toastStyles);
+const GlobalStyle = createGlobalStyle`${toastCss}`;
+
+export const ToastBody: React.FunctionComponent<
+  ICourierToastMessage & {
+    theme: Theme;
+    brand: Brand;
+  }
+> = ({ theme, brand, ...props }) => {
+  props.icon = props.icon ?? brand?.inapp?.icons?.message;
```
suhasdeshpande, line 21: Could help me understand what configs or settings comes from studio design vs directly from the component props and what is the place to make more additions if I were to get more things coming from Studio designer?

packages/storybook/stories/toast/examples.stories.tsx
```diff
@@ -18,20 +18,38 @@ export default {
...
+      <ToastBody
+        body={bodyText}
+        title="Title"
+        brand={{
+          inapp: {
+            colors: {
+              invertButtons: true,
+            },
+            icons: {
+              message:
```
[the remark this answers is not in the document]
rileylnapier: icons.messageIcon?  feels a little redundant no?

packages/react-toast/tsconfig.json
```diff
@@ -8,12 +8,13 @@
...
-    "node_modules",
+    "**/__mocks__/**",
+    "**/__tests__/**",
     "**/*.spec.ts",
     "**/*.story.ts",
     "**/dist/**",
-    "**/__tests__/**",
-    "**/__mocks__/**"
+    "**/typings/**",
+    "node_modules"
```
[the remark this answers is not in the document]
rileylnapier: its so we don't remove the typings folder each time we yarn start which was confusing vscode

packages/react-toast/src/components/index.tsx
rileylnapier: props.icon means that the icon is coming from a websocket which has higher prescedent than brand icon

packages/react-provider/src/hooks/use-courier.ts
```diff
@@ -1,6 +1,6 @@
 import { useContext } from "react";
 import { CourierContext } from "../";
-import { ICourierContext } from "../types";
+import { CourierContext as ICourierContext } from "../types";
```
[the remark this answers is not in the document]
rileylnapier: i kind of feel like we shouldn't use I in front of our types, because I never see that in public packages.  lets chat about it

packages/react-inbox/src/components/Messages/Header/index.tsx
rileylnapier: we shouldn't expose it to yet no.  some people are trying to implement inbox right now and i don't think we are ready for them to see preferences yet right?

packages/react-inbox/src/components/Message/styled.tsx
rileylnapier: yeah we should do that

suhasdeshpande (dismissed):

suhasdeshpande (approved):

packages/react-toast/src/components/Body/styled.ts
```diff
@@ -9,8 +9,8 @@ export const iconStyles = ({ theme }) =>
       objectFit: "contain",
       alignSelf: "center",
       padding: "2px",
-      maxHeight: "35px !important",
-      maxWidth: "35px !important",
+      maxHeight: "40px !important",
```
suhasdeshpande, line 12: is !important really necessary?

rileylnapier: yeah it is,  i removed it at one point at it screwed something up

rileylnapier: i think we can fix it but for now i think its ok

PR #77, opened 2021-05-06
Beta1.1

rileylnapier:
## Description

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-provider/src/index.tsx
```diff
@@ -122,11 +122,15 @@ export const CourierProvider: React.FunctionComponent<ICourierContext> = ({
...
+    if (!graphQLClient.canRequest) {
+      return;
+    }
+
     dispatch({
       type: "root/GET_BRAND",
       payload: () => getBrand(graphQLClient, brandId),
     });
-  }, [brand, brandId]);
+  }, [graphQLClient, brand, brandId]);
```
suhasdeshpande, line 133: would `graphQLClient.canRequest` as deps give more control over when this effect is called?

suhasdeshpande (approved):
small comment around dependency

packages/react-provider/src/index.tsx
rileylnapier: it `could` but it doesn't because this is how the client is created:

```
  const graphQLClient = useMemo(() => {
    return new Client({ clientKey, userId, userSignature, apiUrl });
  }, [clientKey, userId, userSignature, apiUrl]);
```

rileylnapier: so it makes a new one anyway

suhasdeshpande (dismissed):

suhasdeshpande (approved):

PR #90, opened 2021-06-03
blocks

rileylnapier:
## Description

image

image

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-toast/src/components/Body/styled.ts
```diff
@@ -38,15 +73,73 @@ export const Message = styled.div(({ theme }) =>
...
+        },
+
+        "&:hover": {
+          color: "#73819B",
+          background: "rgb(0 0 0 / 10%)",
+        },
+
+        borderRadius: 4,
+      },
+      theme?.message?.actionBlock
```
tk26, line 143: prob don't need a ? after theme

packages/react-toast/src/components/Body/index.tsx
```diff
@@ -49,20 +56,53 @@ const ToastBody: React.FunctionComponent<Partial<ICourierToastMessage>> = ({
...
-        onClickDetails={handleOnClickDetails}
-        onClickDismiss={handleOnClickDismiss}
-      />
-    </>
+    <Container>
+      {Icon && <Icon data-testid="message-icon" />}
+      <Message data-testid="message">
+        {title && <Title data-testid="message-title">{title}</Title>}
+        {blocks?.length ? (
+          blocks?.map((block, index) => {
```
tk26, line 64: don't need ? because length is > 0

packages/react-toast/src/components/Body/index.tsx
```diff
@@ -49,20 +56,53 @@ const ToastBody: React.FunctionComponent<Partial<ICourierToastMessage>> = ({
...
+              );
+            }
+          })
+        ) : (
+          <>
+            <TextBlock data-testid="message-body">{body}</TextBlock>
+            {data?.clickAction && (
+              <ActionBlock
+                data-testid="action-0"
+                href={data?.clickAction}
```
tk26, line 92: don't need `?`

packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -79,10 +81,36 @@ const Message: React.FunctionComponent<IMessageProps> = ({
       {renderedIcon}
       <Contents>
         <Title>{title}</Title>
-        <Body>{body}</Body>
+        {blocks?.length ? (
+          blocks?.map((block, index) => {
```
tk26, line 85: don't need ?

tk26 (approved):
some super minor comments which is nbd

packages/react-toast/src/components/Body/index.tsx
rileylnapier: yeah its habit at this point

PR #103, opened 2021-06-22
courier-react tweaks

rileylnapier:
## Description

- Unread and read notification should be aligned. The unread notification below currently has an indent
- The preferences gear and the "mark all as read" cta should both be #73819B. 
- Change branding so the tabs, notification CTAs and unread indicator are all the primary CTA. The secondary/tertiary branding will be covered by our background gradient in inbox v2
- If possible we should scale the inbox down to 360px width. That's what we have in our figma.
- Notification border radius
- When should the toast notification disappear
- Can we tweak the toggle so it follows this toggle design?
- The alignment in the preference grouping looks off. can we adjust so the margin above the preference category title and below opted in is the same

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-toast/src/components/Toast/index.tsx
```diff
@@ -34,8 +34,16 @@ export const Toast: React.FunctionComponent<
...
+  const {
+    clientKey,
+    transport,
+    dispatch,
+    brand: courierBrand,
+  } = courierContext;
 
+  const brand = config?.brand ?? courierBrand;
+
+  console.log("brand", brand);
```
tk26, line 46: remove

tk26 (dismissed):

packages/react-toast/src/components/Toast/index.tsx
suhasdeshpande:
```suggestion

```

packages/storybook/stories/toast/examples.stories.tsx
```diff
@@ -124,6 +124,9 @@ export function WithCourierProvider({ bodyText }) {
               message:
                 "https://d33wubrfki0l68.cloudfront.net/ca2747f11cc64d0e424e27b4a804b9d981b22453/9ab46/_next/static/images/logo@2x-5d5af82635bfdd3ad24e54f9eb364097.png",
             },
+            toast: {
+              autoCloseTimer: 0,
```
suhasdeshpande, line 128: I was not able to see inbox examples in deploy preview for some reason?

rileylnapier: its because the preview is missing env variables so the story borks

tk26 (approved):

PR #110, opened 2021-06-29
ws reconnect

rileylnapier:
## Description

when we disconnect and reconnect we need to resubscribe so we need to keep track of all the subscriptions

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-provider/package.json
+++ b/packages/react-provider/package.json
@@ -11,7 +11,7 @@
     "clean": "rimraf dist",
     "test": "jest -c jest.config.js --runInBand --silent",
     "type-check": "tsc --noEmit",
-    "types": "tsc --declaration --outDir typings/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false"
+    "types": "tsc --showConfig --declaration --outDir typings/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false"
   },
   "license": "ISC",
   "dependencies": {
--- a/packages/react-provider/package.json
+++ b/packages/react-provider/package.json
@@ -11,7 +11,7 @@
     "clean": "rimraf dist",
     "test": "jest -c jest.config.js --runInBand --silent",
     "type-check": "tsc --noEmit",
-    "types": "tsc --showConfig --declaration --outDir typings/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false"
+    "types": "tsc --declaration --outDir typings/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false"
   },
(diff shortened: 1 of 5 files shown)


packages/react-provider/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "graphql": "^15.5.0",
     "react-use": "^17.2.1",
+    "reconnecting-websocket": "^4.4.0",
```
suhasdeshpande, line 20: this one was last updated 1 year ago?

packages/react-provider/src/ws.ts
```diff
@@ -1,74 +1,91 @@
...
+    this.connection = new ReconnectingWebSocket(
+      `${this.url}/?clientKey=${this.clientKey}`
+    );
+
+    this.connection.onopen = this.onOpen.bind(this);
     this.connection.onmessage = this.onMessage.bind(this);
   }
 
+  onOpen(): void {
+    this.connected = true;
```
tk26, line 35: should we also do `this.connected = false` in `close()`?

rileylnapier: yeah prob, good idea

tk26 (approved):

packages/react-provider/package.json
rileylnapier: looks like it yeah

suhasdeshpande: it also looks slightly heavy. but is it getting tree shaken? I am mostly curious if adding this dependency bloats our package or not.

scarney81: @suhasdeshpande this package is 2.6kb minified and gzipped https://bundlephobia.com/package/reconnecting-websocket@4.4.0

scarney81: It's also tree shakeable and has no dependencies.

Screen Shot 2021-06-30 at 3 28 52 PM

rileylnapier: curious, what made you think it was slighly heavy?   i don't see any indication of that.

tk26 (approved):

packages/react-provider/package.json
suhasdeshpande: I didn't realize I was looking at unpacked size on npm. 
Screen Shot 2021-06-30 at 3 50 35 PM

rileylnapier: ahh, isee

scarney81: Unpacked size can be deceiving. Mobx is 3.84 MB unpacked but like 16.1kB minified and gzipped

PR #114, opened 2021-07-08
new design with gradients

rileylnapier:
## Description

image

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Messages/styled.ts
```diff
@@ -1,14 +1,29 @@
 import styled from "styled-components";
 import deepExtend from "deep-extend";
 
+export const Container = styled.div(({ theme }) =>
+  deepExtend(
+    {
+      background:
+        theme.brand.inapp?.widgetBackground?.topColor &&
```
tk26, line 8: maybe we should do optional chaining on brand too.

tk26 (dismissed):
Minor suggestion

cgradwohl (approved):

packages/react-inbox/src/components/Messages/styled.ts
rileylnapier: yes probably.  good catch

tk26 (approved):

PR #126, opened 2021-10-29
upgrade webpack for angular

rileylnapier:
## Description

drone deploy was having issues with our embedded solution.  it works in my machine but not in angular.  it works in angular if its not minimized.  i upgraded webpack to v5 which uses terser plugin instead of uglify and that seems to fix it

## Type of change

## Related issues

> `Fix [#1]()`


packages/components/webpack.config.js
```diff
@@ -12,9 +12,7 @@ module.exports = (env, argv) => {
     mode: argv.mode ? argv.mode : "development",
     entry: path.resolve(__dirname, "./src/index.tsx"),
     output: {
-      publicPath: isProduction
-        ? "https://courier-components-xvdza5.s3.amazonaws.com/"
-        : undefined,
+      publicPath: isProduction ? "assets/" : "assets/",
```
tk26, line 15: is this ternary intended?

rileylnapier: no, thank you

tk26 (approved):

PR #135, opened 2021-12-02
[C-4547] fixes opacity on "end of the road" graphic in inbox widget

suhasdeshpande:
## Description
Screen Shot 2021-12-01 at 4 05 34 PM

## Type of change


packages/react-inbox/src/components/Messages/styled.ts
```diff
@@ -98,7 +98,7 @@ export const Footer = styled.div(({ theme }) =>
 );
 
 export const Line = styled.div(({ theme }) => ({
-  backgroundColor: theme?.brand?.colors?.primary ?? "#9121c2",
+  backgroundColor: theme?.brand?.colors?.primary ?? "#F9FAFB",
```
tk26, line 101: isn't 9121c2 a default we use at a bunch of places in frontend?

suhasdeshpande: that's not the default we want for background color afaik

suhasdeshpande: good catch. I wasn't intended to change this.

tk26 (dismissed):

rileylnapier (approved):

PR #140, opened 2022-01-07
update courier-react for hmac authorization on subscribe

rileylnapier:
## Description

we need to make sure to pass the user signature in the body to validate against 
## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-provider/src/ws.ts
+++ b/packages/react-provider/src/ws.ts
@@ -1,5 +1,6 @@
 import { ICourierEventCallback, ICourierMessage } from "./transports/types";
 import ReconnectingWebSocket from "reconnecting-websocket";
+import packageJson from "../package.json";
 
 export class WS {
   connection?: ReconnectingWebSocket;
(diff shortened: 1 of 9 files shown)


packages/react-provider/src/ws.ts
```diff
@@ -29,9 +40,15 @@ export class WS {
 
     this.connection.onopen = this.onOpen.bind(this);
     this.connection.onclose = this.onClose.bind(this);
+    this.connection.onerror = this.onError.bind(this);
     this.connection.onmessage = this.onMessage.bind(this);
   }
 
+  onError(): void {
+    console.error("Error Connecting to Courier Push");
```
suhasdeshpande, line 48: can we get a more informative error here?

rileylnapier: lets wait for feedback and optimize premature

NmSand (approved):

tk26 (approved):

PR #163, opened 2022-02-16
Riley/c 5093 ld courier react feedback 0204

rileylnapier:
## Description

add 'renderBlocks' render prop to allow adding/editing custom properties to blocks

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/types.ts
+++ b/packages/react-inbox/src/types.ts
@@ -12,11 +12,10 @@ export interface InboxProps {
   brand?: Brand;
   className?: string;
   defaultIcon?: false | string;
+  from?: number;
   isOpen?: boolean;
   placement?: TippyProps["placement"];
   renderContainer?: React.FunctionComponent;
-  from?: number;
-  showUnreadMessageCount?: boolean;
   renderTabs?: React.FunctionComponent<{
     currentTab?: ITab;
     tabs?: ITab[];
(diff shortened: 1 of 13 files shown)


packages/storybook/stories/inbox/mock-middleware.tsx
```diff
@@ -1,4 +1,5 @@
 export default () => (next) => (action) => {
+  console.log(action);
```
tk26, line 2:
```suggestion
```

packages/storybook/stories/inbox/index.stories.tsx
```diff
@@ -117,6 +121,7 @@ export const RenderPropsExample = () => {
     renderIcon: CustomIcon,
     renderTabs: CustomTabs,
     renderNoMessages: CustomNoMessages,
+    renderMessageBlock: CustomMessagesMessageBlock,
```
tk26, line 124: remove?

packages/storybook/stories/inbox/custom-components.tsx
```diff
@@ -27,3 +27,7 @@ export const CustomNoMessages: React.FunctionComponent = () => {
 export const CustomBell: React.FunctionComponent = () => {
   return <div>Bell</div>;
 };
+
+export const CustomMessagesMessageBlock: React.FunctionComponent = () => {
```
tk26, line 31: this would be just `CustomBlock` no?

packages/storybook/stories/inbox/index.stories.tsx
tk26: we need `renderBlocks` as prop

packages/react-inbox/src/types.ts
```diff
@@ -27,13 +26,18 @@ export interface InboxProps {
     onClick?: (event: React.MouseEvent) => void;
     onMouseEnter?: (event: React.MouseEvent) => void;
   }>;
+  renderBlocks?: {
```
tk26, line 29: can someone send multiple text/action blocks? do we need that to be an array so it isn't a breaking change down the road?

rileylnapier: its a function that will be called for each block...

tk26 (approved):

PR #170, opened 2022-03-10
add client-graphql package

rileylnapier:
## Description

lets expose the graphql api package for use

## Type of change

## Related issues

> `Fix [#1]()`


packages/client-graphql/src/client.ts
```diff
@@ -0,0 +1,39 @@
+import { Client, createClient } from "urql";
```
tk26, line 1: didn't know about `urql` - nice

packages/react-inbox/src/components/Messages/index.tsx
```diff
@@ -110,7 +110,7 @@ const Messages: React.ForwardRefExoticComponent<
                   <Message key={message.messageId} {...message} />
                 )
               )}
-              {isLoading && <Loading />}
+              {isLoading && messages?.length === 0 && <Loading />}
```
tk26, line 113: why do we need to add this?

packages/client-graphql/src/client.ts
rileylnapier: its a smaller package

packages/react-inbox/src/components/Messages/index.tsx
rileylnapier: just to make the inbox nicer.  that if you open, close and reopen it doesn't show the loading screen again...

tk26: oh nice we had a couple of folks asking about that fluctuation

tk26 (approved):

tk26 (approved):

PR #218, opened 2022-07-21
fix brand fetching and re-renders

rileylnapier:
## Description

with async props we need to make sure we fetch the brand and make sure we don't rerender 

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-provider/src/hooks/use-why-did-you-update.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useRef } from "react";
```
rileylnapier, line 1: this is to help with dev

packages/react-provider/src/index.tsx
```diff
@@ -33,155 +33,159 @@ export type {
...
+        // if we pass in brand, don't fetch it
+        return;
+      }
+
+      if (!clientKey || !userId) {
+        return;
+      }
+
+      actions.getBrand(brandId);
+    }, [actions, brand, brandId, clientKey, userId]);
```
rileylnapier, line 146: this useEffect is all that changed, i added conditional on line 141 and updated the deps in 146

suhasdeshpande (approved):

suhasdeshpande (approved):

PR #221, opened 2022-07-25
add new inbox to clientgraphql

rileylnapier:
## Description

this adds new exports to client-graphql for Inbox.  currently we have Messages but Inbox is what will be used for the new inbox apis

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/client-graphql/src/inbox/message.ts
+++ b/packages/client-graphql/src/inbox/message.ts
@@ -1,0 +1,48 @@
+import { Client } from "urql";
+
+export const GET_INBOX_MESSAGE = `
+  query GetInboxMessage($messageId: String!){
+    message(messageId: $messageId) {
+      created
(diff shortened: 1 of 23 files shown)


packages/client-graphql/src/inbox/message.ts
```diff
@@ -0,0 +1,48 @@
...
+      message: any;
+    }
+  | undefined
+>;
+
+export const getInboxMessage =
+  (client?: Client): GetInboxMessage =>
+  async (messageId?: string) => {
+    if (!client) {
+      return Promise.resolve(undefined);
```
drew-y, line 36: If the function is marked async you could safely just `return undefined`.

drew-y (approved):

drew-y (approved):

PR #237, opened 2022-08-04
feedback

rileylnapier:
## Description

- fix mark as read dropdown from not disappearing when you click it 2x
- allow for passing in tabs: [] or tabs: undefined
- fix what happens when you don't supply tabs

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/components/src/index.html
+++ b/packages/components/src/index.html
@@ -37,14 +37,16 @@
       };*/
       window.courierConfig = {
         apiUrl:
-          "https://yubmnstah4.execute-api.us-east-1.amazonaws.com/staging",
-        brandId: "YYBVRK524Y42DEJDMYNMAPNTSCJ6",
-        wsUrl: "wss://icnrz8ttcf.execute-api.us-east-1.amazonaws.com/staging",
-        clientKey: "YWZiZWViNGItMjAyMS00MzgwLTlkZDUtZWI0Y2MzNzEwNmMw",
-        userId: "bc976f74-7978-41c6-9d26-77b630986fd8",
+          "https://rubmz24skk.execute-api.us-east-1.amazonaws.com/dev/client/q",
+        wsOptions: {
+          url: "wss://20en15n3ng.execute-api.us-east-1.amazonaws.com/dev",
+        },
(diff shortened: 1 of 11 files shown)


packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -189,7 +178,7 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
       event.preventDefault();
       if (!isOpen) {
         setView("messages");
-        setCurrentTab(DEFAULT_TABS[0]);
+        setCurrentTab(props?.tabs?.[0] ?? DEFAULT_TABS[1]);
```
rileylnapier, line 181: if no tabs, use ALL MESSAGES for the current tab

packages/components/src/index.html
```diff
@@ -37,14 +37,16 @@
...
-        brandId: "YYBVRK524Y42DEJDMYNMAPNTSCJ6",
-        wsUrl: "wss://icnrz8ttcf.execute-api.us-east-1.amazonaws.com/staging",
-        clientKey: "YWZiZWViNGItMjAyMS00MzgwLTlkZDUtZWI0Y2MzNzEwNmMw",
-        userId: "bc976f74-7978-41c6-9d26-77b630986fd8",
+          "https://rubmz24skk.execute-api.us-east-1.amazonaws.com/dev/client/q",
+        wsOptions: {
+          url: "wss://20en15n3ng.execute-api.us-east-1.amazonaws.com/dev",
+        },
+        clientKey: "NzY4MjUxY2YtM2ViOC00MjZhLTkyZWItZmFhMGU3Njc4NzY4",
+        userId: "70f6a4f4-2907-4518-b8f3-b9cfab224764",
```
tk26, line 45: any reason why we're changing staging to dev?

rileylnapier: this is a development only thing so i changed it to my dev

tk26: gotcha

tk26 (approved):

PR #244, opened 2022-08-10
better fetching for initial state and all tabs

rileylnapier:
## Description

now that we're mving to local state we need to do better at fetching data

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-provider/src/hooks/use-transport.ts
```diff
@@ -0,0 +1,25 @@
+import { useMemo } from "react";
```
rileylnapier, line 1: attempting to breakup some of the bigger components into smaller hooks

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -180,14 +175,25 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
       event.preventDefault();
+      let myCurrentTab = currentTab;
       if (!isOpen) {
         setView("messages");
-        setCurrentTab(tabs?.[0] ?? DEFAULT_TABS[1]);
+        myCurrentTab = tabs?.[0] ?? DEFAULT_TABS[1];
+        setCurrentTab(myCurrentTab);
       }
 
+      if (!lastMessagesFetched) {
```
rileylnapier, line 185: will probably do something with this to "refetch" if a certain time has elapsed but can deal with that in later optimiation

packages/client-graphql/src/__tests__/messages.spec.ts
```diff
@@ -0,0 +1,241 @@
...
...}\\\\n      __typename\\\\n    }\\\\n    __typename\\\\n  }\\\\n  list1: messages(params: $params1, limit: $limit) {\\\\n    totalCount\\\\n    pageInfo {\\\\n      startCursor\\\\n      hasNextPage\\\\n      __typename\\\\n    }\\\\n    nodes {\\\\n      id\\\\n      messageId\\\\n      created\\\\n      read\\\\n      tags\\\\n      content {\\\\n        title\\\\n        body\\\\n        blocks {\\\\n          ... on TextBlock {\\\\n            type\\\\n            text\\\\n            __typename\\\\n          }\\\\n          ... on ActionBlock {\\\\n            type\\\\n            text\\\\n            url\\\\n            __typename\\\\n          }\\\\n          __typename\\\\n        }\\\\n        data\\\\n        trackingIds {\\\\n          archiveTrackingId\\\\n          clickTrackingId\\\\n          deliverTrackingId\\\\n          readTrackingId\\\\n          unreadTrackingId\\\\n          __typename\\\\n        }\\\\n        __typename\\\\n      }\\\\n      __typename\\\\n    }\\\\n    __typename\\\\n  }\\\\n}\\\\n\\",\\"operationName\\":\\"GetMessages\\",\\"variables\\":{\\"params0\\":{\\"isRead\\":false},\\"params1\\":{\\"tags\\":[\\"abc\\",\\"123\\"]},\\"limit\\":10}}",
```
drew-y, line 157: Beautiful

packages/client-graphql/src/brands.ts
```diff
@@ -54,7 +54,7 @@ query GetBrand($brandId: String!) {
 `;
 
 const GET_INAPP_BRAND = `
-query GetInAppBrand {
+query GetInppBrand {
```
drew-y, line 57: Was this change on purpose? I feel like this might be a typo.

packages/client-graphql/src/initial-state.ts
```diff
@@ -0,0 +1,97 @@
...
+
+interface IInitialStateParams {
+  skipFetchBrand?: boolean;
+  brandId?: string;
+  from?: number;
+}
+
+type GetInitialState = (params: IInitialStateParams) => Promise<
+  | {
+      brand: any;
```
drew-y, line 14: Too bad we don't have types for this :/

packages/client-graphql/src/initial-state.ts
```diff
@@ -0,0 +1,97 @@
...
+  | undefined
+>;
+
+const buildQuery = ({
+  skipFetchBrand,
+  brandId,
+}: Partial<IInitialStateParams>) => {
+  if (skipFetchBrand) {
+    return `
+      query InitialState($params: FilterParamsInput) {
```
drew-y, line 26: Is it possible to use `directives` here in place of string interpolation here? Just thinking aloud.

drew-y: Apparently not :/

drew-y: Only supports fields.

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -96,14 +96,17 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
   }, [props]);
 
   const tabs = props.tabs == false ? undefined : props.tabs;
```
drew-y, line 98:
```suggestion
  const tabs = props.tabs === false ? undefined : props.tabs;
```

packages/client-graphql/src/brands.ts
drew-y, line 57:
```suggestion
query GetInAppBrand {
```

drew-y (approved):

PR #267, opened 2022-09-07
Christian/c 6951 create overall footer component in

heyglassy:
## Description

Consolidated Business and free tier footers into one footer.

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-hooks/src/preferences/types.ts
```diff
@@ -4,6 +4,7 @@ export interface PreferenceState {
   preferences?: IPreferenceTemplate[];
   recipientPreferences?: IRecipientPreference[];
   preferenceSections?: Array<any>;
+  preferencePage?: any;
```
suhasdeshpande, line 7: do we need preferenceSections and preference page at the same level?

heyglassy: for now, I'll create another ticket to move deprecate preferenceSections.

packages/react-preferences/src/components/BusinessFooter.tsx
```diff
@@ -31,46 +31,49 @@ const FooterWrapper = styled.div`
...
+}) => {
   return (
     <FooterWrapper>
       <SVGWrapper>
-        <Facbebook />
-        <Instagram />
-        <LinkedIn />
-        <Medium />
-        <Twitter />
+        {links?.facebook?.url && (
```
suhasdeshpande, line 52: This feels a bit awkward datastrucuture-wise. might be easier to map this to a  strongly typed collection on the API side.

```ts
type SocialLinks = Array<{type: "facebook" | "Instagram" | string, value: string}>
```

packages/storybook/stories/inbox/examples.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+            brand={{
+              colors: {
+                primary: "red",
+                secondary: "pink",
+                tertiary: "orange",
+              },
+            }}
+          />
+        </CourierProvider>
+      </div> */}
```
suhasdeshpande, line 198: do we need this?

heyglassy: fixed

suhasdeshpande (approved):

PR #280, opened 2022-09-20
fix hooks and actions

rileylnapier:
## Description

- get unread message count payload type was wrong
- don't fetch "initial state" if we are only using hooks

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-hooks/src/inbox/elemental-inbox/use-inbox-actions.ts
+++ b/packages/react-hooks/src/inbox/elemental-inbox/use-inbox-actions.ts
@@ -77,7 +77,6 @@ const useElementalInboxActions = (): IInboxActions => {
         type: "inbox/FETCH_UNREAD_MESSAGE_COUNT",
         payload: () =>
           inboxClient.getInboxCount({
-            tags: [],
             status: "unread",
           }),
       });
--- a/packages/react-hooks/src/inbox/use-inbox-actions.ts
+++ b/packages/react-hooks/src/inbox/use-inbox-actions.ts
@@ -73,7 +73,7 @@ const useInboxActions = (): IInboxActions => {
       dispatch(initInbox(payload));
 
       const response = await initialState.getInitialState({
-        brandId: brandId,
+        brandId,
         skipFetchBrand: Boolean(brand),
       });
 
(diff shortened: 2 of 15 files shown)


packages/react-hooks/src/inbox/elemental-inbox/use-inbox-actions.ts
```diff
@@ -77,7 +77,6 @@ const useElementalInboxActions = (): IInboxActions => {
         type: "inbox/FETCH_UNREAD_MESSAGE_COUNT",
         payload: () =>
           inboxClient.getInboxCount({
-            tags: [],
```
rileylnapier, line 80: this was pulling back 0 for count

packages/react-hooks/src/inbox/use-inbox-actions.ts
```diff
@@ -56,11 +66,29 @@ const useInboxActions = (): IInboxActions => {
...
   const events = Events({ client: courierClient });
   const messages = Messages({ client: courierClient });
+  const initialState = InitialState({ client: courierClient });
 
   return {
-    init: (payload) => {
+    init: async (payload) => {
       dispatch(initInbox(payload));
 
+      const response = await initialState.getInitialState({
```
rileylnapier, line 75: if someone was only to use the hooks... i feel like we maybe shouldn't fetch initial state right?  we should only do this initial state stuff if we are using the components?

rileylnapier: jk this "init" action only gets fired if you are using the components so it should be fine

packages/storybook/stories/inbox/full-page-inbox.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+          justifyContent: "space-between",
+        }}
+      >
+        <div>
+          <ReactMarkdown>{`## Example`}</ReactMarkdown>
+          <ReactMarkdown>{`\`\`\`javascript\n${fullPageInboxHooksString}\n\`\`\``}</ReactMarkdown>
+        </div>
+        <CourierProvider
+          wsOptions={{
+            url: "wss://icnrz8ttcf.execute-api.us-east-1.amazonaws.com/staging",
```
rileylnapier, line 30: not sure if this matters i left staging on here.  it was how i tested the CPP changes

drew-y (approved):

drew-y (dismissed):

drew-y (approved):

drew-y (approved):

PR #302, opened 2022-10-11
Riley/d 169 final inbox review

rileylnapier:
## Description

css tweaks

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/hooks/use-hover.ts
```diff
@@ -1,18 +1,24 @@
...
+    setValue(shouldHover);
+  };
   const handleMouseOut = () => setValue(false);
   useEffect(
     () => {
       const node = ref.current;
       if (node) {
-        node.addEventListener("mouseenter", handleMouseOver);
+        node.addEventListener("mouseover", handleMouseOver);
         node.addEventListener("mouseleave", handleMouseOut);
```
tk26, line 19: do we need to change this to mouseout? https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event says mouseleave does not bubble up but mouseout does

rileylnapier: idk let me try and see if it still works, i might not even need to change it

rileylnapier: i didn't need to change this

tk26 (approved):

PR #310, opened 2022-10-17
[C-7266] React hooks renew session function + doc updates

drew-y:
## Description

Adds `renewSession` method to the elemental inbox hooks. This allows websocket connections with short lived authorization tokens to renew their sessions indefinitely.

Also includes updates and improvements to related docs.

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-hooks/docs/0.overview.md
+++ b/packages/react-hooks/docs/0.overview.md
@@ -25,8 +25,11 @@ const MyApp = () => {
   /**
    * Auth token for courier provider, can be a token from Courier's auth/issue-token endpoint
    * or a JWT signed with a valid courier api key. Must include scope: "user_id:<user_id_here> inbox:read:messages"
+   *
+   * For more information on the auth/issue-token endpoint, visit:
+   * https://courier.com/docs/reference/auth/intro/
    */
-  const authorization = fetchAuthToken();
+  const authorization = await fetchAuthToken();
 
   return (
     <CourierProvider authorization="abc123">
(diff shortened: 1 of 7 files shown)


packages/react-hooks/docs/0.overview.md
```diff
@@ -42,18 +45,29 @@ const MyInbox = () => {
     inbox.fetchMessages();
   }, []);
 
+  // Updates message.status to "read"
```
rileylnapier, line 48: im not sure "status" here is read/unread.  i think `read` is a timestamp?  but the code returns it as read:boolean?

packages/react-hooks/docs/0.overview.md
```diff
@@ -42,18 +45,29 @@ const MyInbox = () => {
...
 
+  // Archived messages are included in inbox.fetchMessages()
   const handleArchiveMessage = (message) => () => {
     inbox.markMessageArchived(message.messageId);
   };
 
+  // If the supplied authorization token is short lived, renew the session with a fresh token
+  // proactively before the token is set to expire. Here we use 5 minutes assuming our token only
+  // lasts 10 minutes
+  setInterval(() => {
```
rileylnapier, line 66: this would need to be in a useEffect

packages/react-hooks/docs/0.overview.md
```diff
@@ -42,18 +45,29 @@ const MyInbox = () => {
...
   const handleReadMessage = (message) => () => {
     inbox.markMessageRead(message.messageId);
   };
 
+  // Updates message.status to "unread"
   const handleUnreadMessage = (message) => () => {
     inbox.markMessageUnread(message.messageId);
   };
 
+  // Archived messages are included in inbox.fetchMessages()
```
rileylnapier, line 58: archived messages are `not` returned

packages/react-hooks/docs/0.overview.md
drew-y: Ahh. You're right. There are issues with our types here.

packages/react-hooks/docs/0.overview.md
drew-y: Good catch

PR #312, opened 2022-10-19
update inbox fetching and mark all as read

rileylnapier:
## Description

- only fetch new messages if new messages haven't been fetched in the last hour
- save lastmarkedasreaddate so we don't show incorrect date

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-hooks/src/inbox/elemental-inbox/actions/init.ts
```diff
@@ -1,13 +0,0 @@
-import { IElementalInbox } from "../types";
```
rileylnapier, line 1: these were the same as normal inbox actions so they were duplicates

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -161,22 +162,31 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
 
-        if (tabs) {
-          fetchMessageLists(tabs);
-        } else {
-          fetchMessages({
-            params: myCurrentTab.filters,
-          });
+        console.log("lastMessagesFetched", lastMessagesFetched);
+        console.log("dateDiff", dateDiff);
+        if (!dateDiff || dateDiff > 3600000) {
```
rileylnapier, line 176: if no messagesfetched in the last hour, lets refetch

rileylnapier: logic here could be a discussion

drew-y (approved):

PR #314, opened 2022-10-19
fix onclick for unread messages

rileylnapier:
## Description

make sure to update localstate

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -72,6 +72,12 @@ const Message: React.FunctionComponent<
     unreadTrackingId,
   });
 
+  const handleMarkAsReadOnClick = () => {
+    if (!read) {
+      markMessageRead(messageId);
```
rileylnapier, line 77: if someone has CTT disabled this may cause confusion

tk26 (approved):

PR #318, opened 2022-10-24
make entire toast clickable

rileylnapier:
## Description

- we're moving away from buttons and want to support clicking on the entire message
- only fire "deliverfed" from a toast vs the provider

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Messages2.0/Message.tsx
```diff
@@ -175,7 +175,7 @@ const MessageWrapper: React.FunctionComponent<
     if (!read && trackingIds?.clickTrackingId && trackingIds?.readTrackingId) {
```
drew-y, line 175: Do we need to update this check?

drew-y: Never mind. Looks like the answer is no.

drew-y (approved):

PR #323, opened 2022-11-03
open tracking

rileylnapier:
## Description

send open tracking for messages rendered

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Message/index.tsx
```diff
@@ -78,6 +79,14 @@ const Message: React.FunctionComponent<
     }
   };
 
+  useEffect(() => {
```
rileylnapier, line 82: might be better to use a "in view" hook vs "rendered"

packages/react-inbox/src/hooks/use-on-screen.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from "react";
+
+export const useOnScreen = (ref, cb) => {
+  useEffect(() => {
```
suhasdeshpande, line 4: would naming function would aid developer experience?

packages/react-toast/src/components/Toast/index.tsx
```diff
@@ -49,9 +49,21 @@ export const Toast: React.FunctionComponent<
...
           : message;
 
-      toast(<Body {...message} icon={message?.icon ?? config?.defaultIcon} />, {
-        role: config?.role ?? "status",
-      });
+      toast(
+        <Body
+          {...message}
+          icon={
+            typeof message?.icon === "string"
```
suhasdeshpande, line 56: why can't we callback to either `message?.icon || config.defaultIcon`? This is a bit hard to read and I am not sure if this is a by-product of some product specification.

suhasdeshpande (approved):
overall fine. I am a bit foggy on some logic and it is verbose.

packages/react-toast/src/components/Toast/index.tsx
rileylnapier: the type made me do it like this, you are right

PR #332, opened 2022-11-21
Christian/c 7680 add draft mode to preference page in

heyglassy:
## Description

In order for the preview preference page to be updated without publishing the page in the preference center, the preference page needs a draft mode. Draft mode enables the preference page to pull from the "drafts" of the preference page instead of the published preference page.

https://www.loom.com/share/a4e340f975db41dea0a66e0178c5ef89

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/components/src/components/PreferencePage.tsx
+++ b/packages/components/src/components/PreferencePage.tsx
@@ -1,12 +1,14 @@
 import { Footer, Header, PreferenceList } from "@trycourier/react-preferences";
 import React from "react";
 
-const PreferencePage: React.FunctionComponent = () => {
+const PreferencePage: React.FunctionComponent<{ draft?: boolean }> = ({
+  draft = false,
+}) => {
   return (
     <div>
-      <Header />
-      <PreferenceList />
-      <Footer />
+      <Header draft={draft} />
+      <PreferenceList draft={draft} />
+      <Footer draft={draft} />
     </div>
   );
 };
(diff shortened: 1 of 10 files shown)


packages/components/src/components/PreferencePage.tsx
```diff
@@ -1,12 +1,14 @@
...
+  draft = false,
+}) => {
   return (
     <div>
-      <Header />
-      <PreferenceList />
-      <Footer />
+      <Header draft={draft} />
+      <PreferenceList draft={draft} />
+      <Footer draft={draft} />
```
suhasdeshpande, line 11: how's draft changing footer and header? or you are just making sure you add support for it in future?

packages/components/src/components/index.tsx
```diff
@@ -18,6 +18,8 @@ const querySelector = (element: HTMLElement, selector: string) => {
 };
 
 export const CourierComponents: React.FunctionComponent = () => {
+  const preferencePageDraftMode =
+    window.courierConfig?.preferencePageDraftMode ?? false;
```
suhasdeshpande, line 22: we should also accept this attribute as prop in case a user wants to use this as pure react component.

packages/components/src/components/PreferencePage.tsx
heyglassy: The header and footer have a use the `showCourierFooter` field from the preferences hook so they need the draft field.

packages/components/src/components/index.tsx
heyglassy: I have this supported for the PreferencePage component. The `CourierComponent` I thought was only for when you're using a script tag.

suhasdeshpande (approved):

PR #335, opened 2022-11-23
Riley/c 7711 keep inbox fullpage in sync accross

rileylnapier:
## Description

- handle incoming events for "read/unread" in new inbox apis
- make sure to has the clientSourceId if authorization is passed instead of clientkey

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-hooks/src/inbox/elemental-inbox/use-inbox-actions.ts
+++ b/packages/react-hooks/src/inbox/elemental-inbox/use-inbox-actions.ts
@@ -1,4 +1,8 @@
-import { CourierTransport, useCourier } from "@trycourier/react-provider";
+import {
+  CourierTransport,
+  IInboxMessagePreview,
+  useCourier,
+} from "@trycourier/react-provider";
 import { IElementalInbox } from "./types";
 import { createCourierClient, Inbox } from "@trycourier/client-graphql";
 import { IGetInboxMessagesParams } from "@trycourier/client-graphql";
@@ -10,6 +14,8 @@ import { markAllRead } from "./actions/mark-all-read";
 import { markMessageUnread } from "../actions/mark-message-unread";
 import { markMessageArchived } from "../actions/mark-message-archived";
 import { IInbox } from "../types";
+import { newMessage } from "./actions/new-message";
+import { resetLastFetched } from "../actions/reset-last-fetched";
 
 export interface IFetchMessagesParams {
   params?: IGetInboxMessagesParams;
@@ -23,13 +29,21 @@ interface IInboxActions {
   getUnreadMessageCount: (params?: IGetInboxMessagesParams) => void;
   init: (inbox: IInbox<IElementalInbox>) => void;
   /** Marks all messages as read by setting message.read to the current ISO 8601 date */
-  markAllAsRead: () => void;
+  markAllAsRead: (fromWSEvent?: boolean) => void;
   /** Archives the supplied message, archived messages are not returned by fetchMessages */
-  markMessageArchived: (messageId: string) => Promise<void>;
+  markMessageArchived: (
+    messageId: string,
+    fromWSEvent?: boolean
+  ) => Promise<void>;
   /** Sets message.read to the current ISO 8601 date  */
-  markMessageRead: (messageId: string) => Promise<void>;
+  markMessageRead: (messageId: string, fromWSEvent?: boolean) => Promise<void>;
   /** Removes message.read, signalling that the message is no longer read */
-  markMessageUnread: (messageId: string) => Promise<void>;
+  markMessageUnread: (
+    messageId: string,
+    fromWSEvent?: boolean
+  ) => Promise<void>;
+  newMessage: (transportMessage: IInboxMessagePreview) => void;
+  resetLastFetched: () => void;
   setView: (view: "messages" | "preferences") => void;
   toggleInbox: (isOpen?: boolean) => void;
   /**
@@ -81,13 +95,23 @@ const useElementalInboxActions = (): IInboxActions => {
         });
       }
     },
-    toggleInbox: (isOpen?: boolean) => {
+    resetLastFetched: () => {
+      dispatch(resetLastFetched());
+      dispatch({
+        type: "inbox/FETCH_UNREAD_MESSAGE_COUNT",
+        payload: () =>
+          inboxClient.getInboxCount({
+            status: "unread",
+          }),
+      });
+    },
+    toggleInbox: (isOpen) => {
       dispatch(toggleInbox(isOpen));
     },
     setView: (view: "messages" | "preferences") => {
       dispatch(setView(view));
     },
-    fetchMessages: (payload?: IFetchMessagesParams) => {
+    fetchMessages: (payload) => {
       const params = {
         from: inbox?.from,
         ...payload?.params,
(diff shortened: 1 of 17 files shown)


packages/react-provider/src/transports/types.ts
```diff
@@ -33,16 +34,32 @@ export interface ICourierMessage {
     trackingUrl?: string;
     clickAction?: string;
   };
+  error?: string;
```
suhasdeshpande, line 37: are these optional to support backward compatibility?

rileylnapier: error is optional because it doesn't exist if there is no error

packages/react-hooks/src/inbox/elemental-inbox/use-inbox-actions.ts
```diff
@@ -107,27 +131,46 @@ const useElementalInboxActions = (): IInboxActions => {
...
       dispatch(markMessageRead(messageId));
+      if (fromWSEvent) {
+        return;
+      }
       await inboxClient.markRead(messageId);
     },
-    markAllAsRead: async () => {
+    markAllAsRead: async (fromWSEvent) => {
       dispatch(markAllRead());
+      if (fromWSEvent) {
```
suhasdeshpande, line 143: What's logic here?

packages/react-hooks/src/inbox/use-inbox.ts
```diff
@@ -27,15 +31,15 @@ const useInbox = () => {
...
+        actions.newMessage(courierEvent?.data as ICourierMessage);
       },
     });
 
     transport?.listen({
       id: "event-listener",
       type: "event",
       listener: (courierEvent) => {
-        const data = courierEvent?.data;
+        const data = courierEvent?.data as ICourierEventMessage;
```
suhasdeshpande, line 42: Do we need any validation before we cast this?

packages/react-provider/src/hooks/use-client-source-id.ts
```diff
@@ -0,0 +1,37 @@
...
+  authorization,
+  id,
+  localStorage,
+}): string => {
+  const clientSourceKey = useHash(
+    authorization ? authorization : `${clientKey}/${userId}`
+  );
+
+  return useMemo(() => {
+    if (!localStorage) {
```
suhasdeshpande, line 17: If localstorage is required, why would we need this check?

packages/react-provider/src/hooks/use-client-source-id.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (localStorageClientSourceId) {
+      return id
+        ? `${localStorageClientSourceId}/${id}`
+        : localStorageClientSourceId;
+    }
+
+    const newClientSourceId = uuid.v4();
+    localStorage.setItem(clientSourceIdLSKey, newClientSourceId);
+    return id ? `${newClientSourceId}/${id}` : newClientSourceId;
+  }, [localStorage, clientSourceKey, id]);
```
suhasdeshpande, line 34: The idea is this id will be constant across same session/tab right?

packages/react-provider/src/hooks/use-hash.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from "react";
+
+function hash(text) {
```
suhasdeshpande, line 3: Do we still need to attribute source and inspiration for this?

packages/react-provider/src/transports/types.ts
```diff
@@ -9,15 +9,16 @@ export interface IActionBlock {
   url: string;
   openInNewTab?: boolean;
 }
-export interface ICourierMessage {
-  event?: string;
-  type?: "message" | "event";
-  error?: string;
+
+export interface ICourierEventMessage {
+  event?: "read" | "unread" | "archive" | "mark-all-read";
```
suhasdeshpande, line 14: Why optional?

suhasdeshpande (dismissed):
Overall no objections, i do have some questions overall

packages/react-provider/src/hooks/use-client-source-id.ts
rileylnapier: its not required

packages/react-provider/src/hooks/use-client-source-id.ts
rileylnapier: yes

packages/react-provider/src/hooks/use-hash.ts
rileylnapier: like make a comment?

packages/react-provider/src/transports/types.ts
rileylnapier: i guess its not anymore but events used to not have this so its recently added.

packages/react-hooks/src/inbox/use-inbox.ts
rileylnapier: the type "event" means the type should be ICourierEventMessage.  its possible we have different listener functions that already have the type case so transport.eventListener

packages/storybook/stories/inbox/full-page-inbox.stories.tsx
```diff
@@ -23,22 +23,33 @@ const FramedInBbox = () => {
...
     wsOptions: {
       url: "wss://icnrz8ttcf.execute-api.us-east-1.amazonaws.com/staging",
     },
     apiUrl: "https://4rq7n8hhjd.execute-api.us-east-1.amazonaws.com/staging/q",
-    clientKey: "YWZiZWViNGItMjAyMS00MzgwLTlkZDUtZWI0Y2MzNzEwNmMw",
-    userId: "riley",
-  };
+    clientKey: "MWIxZTk1M2YtMGViYi00N2FkLWJlYWItZjU3OWJkNjA2ZWIx",
+    authorization:
+      "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzY29wZSI6InVzZXJfaWQ6c21va2V5MTIzNDUgaW5ib3g6cmVhZDptZXNzYWdlcyBpbmJveDp3cml0ZTpldmVudHMiLCJ0ZW5hbnRfc2NvcGUiOiJwdWJsaXNoZWQvcHJvZHVjdGlvbiIsInRlbmFudF9pZCI6IjFiMWU5NTNmLTBlYmItNDdhZC1iZWFiLWY1NzliZDYwNmViMSIsImlhdCI6MTY2OTY0OTA1MCwiZXhwIjoxNjY5ODIxODUwLCJqdGkiOiI2YThlZDdkNi0yMjliLTQ2MjktOWNkYy02OGZiOWM2ZGQ1ODAifQ.fYP2nB2PA4ltWukVPR0IXfgudHCTkbMXxvABiDI9TGg",
```
rileylnapier, line 33: this token is expired

packages/react-provider/src/hooks/use-hash.ts
suhasdeshpande: yes

packages/storybook/stories/inbox/full-page-inbox.stories.tsx
suhasdeshpande: why bother keeping it here?

suhasdeshpande (approved):

PR #350, opened 2022-12-12
[C-7830] Hide business/dev footer if no links are provided

NmSand:
## Description

Improve logic to not show business footer if no links are provided
Screen Shot 2022-12-12 at 3 12 12 PM

## Type of change

## Test Plan
- [x] ensure business footer with no links is not displayed

## Related issues

> `Fix [#1]()`


packages/react-preferences/src/components/BusinessFooter.tsx
```diff
@@ -35,8 +35,13 @@ export const BusinessFooter: React.FunctionComponent<{
   links: any;
   theme: string;
 }> = ({ links, theme }) => {
-  if (Object.keys(links).length === 0) {
-    return null;
+  if (
+    typeof links !== "object" ||
```
heyglassy, line 39: This looks good, but the specific case we're trying to prevent is if a link exists but is has an empty string.

Once theres handling around that we should be good to go

`
{
   "links":{
      "facebook":{
         "url":""
      }
   }
}
`

heyglassy (changes requested):
Looks good, see above comment and then this is good to go

heyglassy (approved):

PR #353, opened 2022-12-14
memoize wsOptions, make sure renewSession updates state

rileylnapier:
## Description

- if you pass wsOptions it can cause issues with rerendering so we should memoize it for the consumer
- when you call renewSession,make sure we update state so the next graphql call we make has the right key

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/storybook/stories/inbox/full-page-inbox.stories.tsx
+++ b/packages/storybook/stories/inbox/full-page-inbox.stories.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import ReactMarkdown from "react-markdown";
 
 import { CourierProvider } from "@trycourier/react-provider";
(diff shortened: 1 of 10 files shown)


packages/storybook/stories/inbox/full-page-inbox.stories.tsx
```diff
@@ -65,14 +79,14 @@ export const Hooks = () => {
           <ReactMarkdown>{`## Example`}</ReactMarkdown>
           <ReactMarkdown>{`\`\`\`javascript\n${fullPageInboxHooksString}\n\`\`\``}</ReactMarkdown>
         </div>
-        <CourierProvider {...stagingCourierProps}>
+        <CourierProvider {...devCourierProops}>
           <FullPageInboxHooks />
         </CourierProvider>
-        <Frame>
+        {/*<Frame>
```
tk26, line 85: just making sure commenting this out is intended

rileylnapier: it doesn't really matter as its just the dev enviroment and nobody looks at the storybook

PR #358, opened 2022-12-29
websocket connection updates

rileylnapier:
## Description
a few updates:

1.  autocall renewSession if we see that the new authtoken is for the same scope
2. make sure websocket uses new authtoken on reconnection
3. call keepAlive every 5 minutes

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-provider/src/hooks/use-client-source-id.ts
```diff
@@ -12,9 +9,22 @@ const useClientSourceId = ({
...
+  const clientSourceKey = useMemo(() => {
+    if (!authorization) {
+      return `${clientKey}/${userId}`;
+    }
+
+    const decoded = jwtDecode(authorization) as {
+      tenantId: string;
+      scope: string;
+    };
+    const scopeUserId = decoded?.scope
```
rileylnapier, line 21: not sure how this would work with multiple userids

rileylnapier: but we don't have anyone doing that yet...  i think it opens a can of works to conceptualize

suhasdeshpande (approved):

PR #384, opened 2023-02-03
fix handler

rileylnapier:
## Description

make sure the onReconnect handler replaces the previous function if we are called again with the same id

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-provider/src/ws.ts
```diff
@@ -224,7 +224,13 @@ export class WS {
...
   onReconnection(handler: { id: string; callback: () => void }): void {
     if (this.onReconnectionHandlers.find((h) => h.id === handler.id)) {
-      return;
+      this.onReconnectionHandlers = this.onReconnectionHandlers?.map((h) => {
+        if (h.id === handler.id) {
+          return handler;
+        }
+
+        return h;
+      });
```
tk26, line 233: won't the find give out the h and handler matches anyway? do we even need the if?

rileylnapier: the if is if it isn't already in the array we need to push it

tk26: coding is hard

tk26 (approved):

PR #388, opened 2023-02-13
[C-8196] puts client-version in allow list of custom headers

suhasdeshpande:
## Description

Allow for API to send us the package version so that we can better triage issues and apply targetted patches. 

## Type of change

- [x] chores

### DIFF
--- a/packages/client-graphql/src/client.ts
+++ b/packages/client-graphql/src/client.ts
@@ -1,4 +1,5 @@
 require("isomorphic-fetch");
+import pkg from "../package.json";
 
 import { Client, createClient } from "urql";
 import {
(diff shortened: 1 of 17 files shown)


packages/client-graphql/src/client.ts
```diff
@@ -58,6 +59,13 @@ export const createCourierClient = (
     }
   }
 
+  // add client version and platform
+  headers = {
+    ...headers,
+    "x-courier-client-version": pkg.version,
+    "x-courier-client-platform": "web",
```
rileylnapier, line 66: hmmm, im not sure about the "web" platform here.

suhasdeshpande: any suggestion?

rileylnapier (approved):

PR #390, opened 2023-02-14
Riley/c 8312 make inbox20 designs use new inbox20

rileylnapier:
## Description

switch useInbox hook to use the new apis for fetching messages and tracking events

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-hooks/src/inbox/reducer.ts
+++ b/packages/react-hooks/src/inbox/reducer.ts
@@ -1,4 +1,3 @@
-import { IGraphMessageResponse } from "@trycourier/client-graphql";
 import { IMessage } from "@trycourier/react-provider";
 import { IInbox, ITab } from "./types";
 
--- a/packages/react-hooks/src/inbox/reducer.ts
+++ b/packages/react-hooks/src/inbox/reducer.ts
@@ -113,6 +113,7 @@ type InboxAction =
   | ToggleInbox;
 
 export default (state: IInbox = initialState, action?: InboxAction): IInbox => {
+  console.log("action", action);
   switch (action?.type) {
     case INBOX_INIT: {
       return {
@@ -156,7 +157,7 @@ export default (state: IInbox = initialState, action?: InboxAction): IInbox => {
     case INBOX_FETCH_UNREAD_MESSAGE_COUNT_DONE: {
       return {
         ...state,
-        unreadMessageCount: action.payload,
+        unreadMessageCount: action.payload.count,
       };
     }
 
--- a/packages/react-hooks/src/inbox/reducer.ts
+++ b/packages/react-hooks/src/inbox/reducer.ts
@@ -64,13 +64,23 @@ import {
 } from "./actions/set-current-tab";
 
 export const mapMessage = (
-  message: IGraphMessageResponse,
+  message: any,
   lastMarkedAllRead?: number
 ): IMessage => ({
-  blocks: message.content.blocks,
-  body: message.content.body,
+  blocks: [
+    {
+      type: "text",
+      text: message.preview,
+    },
+    ...message.actions.map((a) => ({
+      type: "action",
+      text: a.content,
+      url: a.href,
+    })),
+  ],
+  body: message.preview,
   created: message.created,
-  data: message.content.data,
+  data: message.data,
   messageId: message.messageId,
   opened: message.opened,
   read: lastMarkedAllRead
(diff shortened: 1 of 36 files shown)


packages/react-hooks/src/inbox/reducer.ts
```diff
@@ -62,24 +61,35 @@ import {
...
   opened: message.opened,
   read: lastMarkedAllRead
     ? lastMarkedAllRead > new Date(message.created).getTime()
       ? true
       : message.read
     : message.read,
-  title: message.content.title,
-  trackingIds: message.content.trackingIds,
+  title: message.title,
+  //trackingIds: message.content.trackingIds,
```
suhasdeshpande, line 92: need this?

packages/react-hooks/src/inbox/use-inbox-actions.ts
```diff
@@ -60,37 +53,66 @@ interface IInboxActions {
...
+      from?: number;
+      status?: "read" | "unread";
+    } = {
+      from,
+      status:
+        tab.filters.isRead === false
+          ? "unread"
+          : tab.filters.isRead === true
+          ? "read"
+          : undefined,
```
suhasdeshpande, line 68: what is this case?

packages/react-inbox/src/components/Messages2.0/Message.tsx
```diff
@@ -184,6 +187,14 @@ const MessageWrapper: React.FunctionComponent<
     }
   };
 
+  useOnScreen(messageHoverRef, () => {
+    if (opened || !messageId) {
+      return;
+    }
+
+    markMessageOpened(messageId);
```
suhasdeshpande, line 195: how do we handle failures or error state here?

packages/react-inbox/src/components/Messages2.0/actions/index.tsx
```diff
@@ -78,32 +77,23 @@ const MessageActions: React.FunctionComponent<{
 
   const handleEvent = (eventName: string) => (event: React.MouseEvent) => {
     event?.preventDefault();
+    if (!messageId) {
```
suhasdeshpande, line 80: what are situations where we won't have messageId?

suhasdeshpande (approved):

packages/react-hooks/src/inbox/reducer.ts
rileylnapier: im not sure exactly tbh, the new inbox apis have all the tracking ids in the db so we don't need to pass them back

packages/react-hooks/src/inbox/use-inbox-actions.ts
rileylnapier: the filter previously was "isRead": true | false and now its "status": "read" | "unread" so making the mapping between the two kinda sucks

packages/react-inbox/src/components/Messages2.0/Message.tsx
rileylnapier: we don't really handle api failures at all.  i wonder what we could do better here

packages/react-inbox/src/components/Messages2.0/actions/index.tsx
rileylnapier: i think we should always have it, let me verify

packages/react-hooks/src/inbox/use-inbox-actions.ts
```diff
@@ -162,28 +189,25 @@ const useInboxActions = (): IInboxActions => {
       dispatch(setCurrentTab(newTab));
     },
     fetchMessages: (payload?: IFetchMessagesParams) => {
+      const searchParams: IGetInboxMessagesParams = {
```
rileylnapier, line 192: make sure this handles status

packages/react-hooks/src/inbox/use-inbox-actions.ts
```diff
@@ -45,13 +41,10 @@ interface IInboxActions {
   getUnreadMessageCount: (params?: IGetMessagesParams) => void;
   init: (inbox: IInbox) => void;
   markAllAsRead: () => void;
-  markMessageArchived: (
```
rileylnapier, line 48: make this not breaking change

suhasdeshpande (approved):

PR #392, opened 2023-02-22
V62 + independent versioning

rileylnapier:
## Description

lets try independent versioning

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "1.61.0",
+  "version": "independent",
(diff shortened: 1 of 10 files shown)


heyglassy (approved):

lerna.json
```diff
@@ -1,6 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "1.61.0",
```
suhasdeshpande, line 3: do we need to update any contribution doc on how to version after this merge?

rileylnapier: same command is run

PR #407, opened 2023-03-09
Fix list unsubscribe text showing for all unsubscribes.

heyglassy:
## Description

Text that indicates a user is unsubscribing from a list appears for all unsubscribes, this issue fixes that.

## Type of change

## Related issues

> `Fix [#1]()`


packages/components/src/index.tsx
```diff
@@ -44,7 +44,7 @@ interface ICourierConfig {
   unsubscribePage?: {
     topicId: string;
     preferencePageUrl: string;
-    list?: boolean;
+    list?: string;
```
misabel, line 47: how does this fix it? @heyglassy

heyglassy: that property is returned as a string from the backend, this will check if it equals the string version of true or false and then pass the boolean to the component.

misabel: so the issue was occurring for the hosted page because it thought it was always a list by default? or?

heyglassy: yes it thought it was list by default because it assumed the backend would pass it a boolean. Because it passed a string, the list was always set to true.

misabel (approved):

PR #423, opened 2023-04-13
feat(click tracking): make click tracking work better by going to the…

rileylnapier:
… new stack

we were piggybacking off the old click redirect tracking and that is slow, this shold make things faster

## Description

- support and subscribe to new action format that has trackingId in `data` vs using redirect urls for both inbox and toast
- deprecate rehydrate messages action in favor of using "init"
- don't refetch message count within an hour to help cache

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-hooks/src/inbox/__tests__/reducer.spec.ts
```diff
@@ -118,26 +114,6 @@ describe("inbox reducer", () => {
     });
   });
 
-  describe(`action ${INBOX_REHYDRATE_MESSAGES}`, () => {
```
rileylnapier, line 121: we just rely on the "init" action to rehydrate

packages/react-hooks/src/inbox/middleware.ts
```diff
@@ -10,7 +10,7 @@ export default (api) => (store) => (next) => async (action) => {
   const state = store.getState();
 
   switch (action.type) {
-    case "root/PAGE_VISIBLE":
+    //case "root/PAGE_VISIBLE":
```
rileylnapier, line 13: i think the page visible, which was leaving a tab and coming back may have caused more issues that it solved

packages/react-hooks/src/inbox/use-inbox-actions.ts
```diff
@@ -102,9 +99,19 @@ const useInboxActions = (): IInboxActions => {
...
+    if (!fromWS) {
+      await inboxClient.markRead(messageId);
+    }
+  };
+
   const handleInit: IInboxActions["init"] = async (payload) => {
     dispatch(initInbox(payload));
-    handleGetUnreadMessageCount();
+    const dateDiff = getDateDiff(payload.lastMessagesFetched);
+    if (!dateDiff || dateDiff > 3600000) {
```
rileylnapier, line 112: only refetch unread count on reload if its been an hour

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -132,28 +132,28 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
   };
 
+  const localStorageState = useLocalStorageMessages(
+    courierContext.clientKey,
+    courierContext.userId
+  );
+
   useEffect(() => {
     init({
+      ...localStorageState,
```
rileylnapier, line 142: this replaced "rehydrate state"

packages/react-hooks/src/inbox/use-inbox-actions.ts
drew-y: Could `dateDiff` be zero? If so, `!dateDiff` would eval to `true`. Which, based on the `dateDiff > 3600000` might be undesirable.

Might be safer to do:
```suggestion
    if (dateDiff === undefined || dateDiff > 3600000) {
```

packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -132,28 +132,28 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
...
     if (!isOpen) {
-      const now = new Date().getTime();
-      const dateDiff = lastMessagesFetched
-        ? now - lastMessagesFetched
-        : undefined;
-
       setView("messages");
 
+      const dateDiff = getDateDiff(lastMessagesFetched);
       if (!dateDiff || dateDiff > 3600000) {
```
drew-y, line 157: Here too.

packages/react-inbox/src/components/Messages2.0/Message.tsx
```diff
@@ -270,20 +260,43 @@ const MessageWrapper: React.FunctionComponent<
...
-        };
+        window.open(clickActionDetails?.href, "_blank");
+      } else {
+        window.location.href = clickActionDetails?.href;
       }
-    }
+    };
 
-    containerProps.onMouseDown = handleClickMessage;
+    //containerProps.onMouseDown = handleClickMessage;
```
drew-y, line 299: Should this still be commented out?

drew-y (approved):
A few non-blocking questions / comments. Looks good!

drew-y (dismissed):

drew-y (approved):

PR #443, opened 2023-05-17
feat: add 3 dot menu topdown for inbox when in mobile mode

heyglassy:
## Description

Add 3 dot menu topdown for inbox when in mobile mode

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/components/Messages2.0/actions/index.tsx
+++ b/packages/react-inbox/src/components/Messages2.0/actions/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import classNames from "classnames";
 
 import { useInbox } from "@trycourier/react-hooks";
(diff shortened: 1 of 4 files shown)


packages/react-inbox/src/components/Messages2.0/actions/index.tsx
```diff
@@ -139,26 +189,66 @@ const MessageActions: React.FunctionComponent<{
...
-          {!read && (
-            <MarkRead
-              label={labels?.markAsRead}
-              onClick={handleEvent("read")}
+        {isMobile ? (
+          <div ref={buttonRef}>
+            <StyledTippy
+              {...tippyProps}
+              content={
+                <div
```
rileylnapier, line 197: can we make sure these are targetable via the theme?  

also can we pass in a renderProp called "renderActionMenu" or something so if someone wants to change the rendering they can?

heyglassy: Is the styled tippy not targetable by theme? I'm using the same thing that is used in a bunch of other places, even though there is a default set of props I'm passing in I thought it would get overridden by the theme

PR #460, opened 2023-06-22
Riley prefs

rileylnapier:
## Description

- remove unnecessary css from preferences so it can import from the main theme
- support inbox as a channel to select in preferences

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/components/src/index.html
+++ b/packages/components/src/index.html
@@ -112,6 +112,9 @@
       const ZedCourierInboxTheme = {
         brand: courierBrand,
         container: {
+          label: {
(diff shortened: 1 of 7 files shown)


packages/components/src/index.html
```diff
@@ -13,6 +13,234 @@
       }
     </style>
     <script type="text/javascript">
+      const fontWeights = {
```
bblatnick32, line 16: @rileylnapier This is just a test file right?

bblatnick32 (approved):

PR #474, opened 2023-07-24
feat: allows filtering inbox by accountId

suhasdeshpande:
## Description

Allows specifying accountId in `courier-provider` 

```tsx
<CourierProvider
  clientKey="<your_client_key>"
  userId="<your_user_id>"
  accountId="<your_account_id>"
>
  <Inbox theme={theme} {...props} />
</CourierProvider>
```

This inbox will fetch and render all messages that were sent to user+accountId. If you don't specify accountId, your inbox will receive all messages sent to the user. 

## Type of change

### DIFF
--- a/packages/react-hooks/src/inbox/use-inbox.ts
+++ b/packages/react-hooks/src/inbox/use-inbox.ts
@@ -23,7 +23,7 @@ const useInbox = (): IInbox<IInboxMessagePreview> & IInboxActions => {
   }
 
   if (accountId) {
-    inbox.accountId = accountId ?? inbox.accountId;
+    inbox.accountId = accountId;
   }
 
   useEffect(() => {
--- a/packages/react-hooks/src/inbox/use-inbox.ts
+++ b/packages/react-hooks/src/inbox/use-inbox.ts
@@ -11,7 +11,7 @@ import useInboxActions, { IInboxActions } from "./use-inbox-actions";
 import { IInbox } from "./types";
 
 const useInbox = (): IInbox<IInboxMessagePreview> & IInboxActions => {
-  const { dispatch, inbox, transport, brand } =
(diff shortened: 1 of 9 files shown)


packages/react-hooks/src/inbox/use-inbox.ts
```diff
@@ -22,6 +22,10 @@ const useInbox = (): IInbox<IInboxMessagePreview> & IInboxActions => {
     inbox.brand = deepExtend({}, brand ?? {}, inbox.brand ?? {});
   }
 
+  if (accountId) {
+    inbox.accountId = accountId ?? inbox.accountId;
```
tk26, line 26: why `??`?

packages/react-provider/src/transports/types.ts
```diff
@@ -1,5 +1,5 @@
 export interface ICourierEventMessage {
-  event: "read" | "unread" | "archive" | "mark-all-read" | "opened";
+  event: "read" | "unread" | "archive" | "mark-all-read" | "opened" | "unpin";
```
tk26, line 2: why we need new event with this change?

drew-y (approved):

packages/react-provider/src/transports/types.ts
suhasdeshpande: the type was missing earlier.

packages/react-hooks/src/inbox/use-inbox.ts
```diff
@@ -22,6 +22,10 @@ const useInbox = (): IInbox<IInboxMessagePreview> & IInboxActions => {
...
+  }
```
suhasdeshpande, line 27:
```suggestion
  if (accountId) {
    inbox.accountId = accountId;
  }
```

rileylnapier (approved):

PR #486, opened 2023-08-01
chore(cleanup storybook): cleanup storybook

rileylnapier:
## Description

- make sure storybook can access env vars from netlify
- make clientkey and auth optional and throw a warning
- fix some type errors storybook was causing

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-inbox/src/components/Messages2.0/Header.tsx
```diff
@@ -322,12 +322,12 @@ const Header: React.FunctionComponent<IHeaderProps> = ({
...
       )}
       <div className="actions">
-        {messages.length > 0 && unreadMessageCount && (
+        {messages.length > 0 && unreadMessageCount ? (
           <MarkAllRead
             label={labels?.markAllAsRead}
             onClick={() => markAllAsRead()}
           />
-        )}
+        ) : null}
```
joshshowalter, line 330: what is the need for this?

packages/react-provider/src/hooks/use-transport.ts
```diff
@@ -55,11 +55,17 @@ const useTransport = ({
     }
 
     if (!clientKey && !authorization) {
-      throw new Error("Missing ClientKey or Authorization");
```
joshshowalter, line 58: we don't need to throw anymore?

packages/react-inbox/src/components/Messages2.0/Message.tsx
```diff
@@ -44,7 +44,7 @@ const UnreadIndicator = styled.div<{ read?: IInboxMessagePreview["read"] }>(
 );
 
 const ClickableContainer = styled.a(({ theme }) => {
-  const primaryColor = theme.brand?.colors?.primary;
+  const primaryColor = theme.brand?.colors?.primary ?? "#9121c2";
```
joshshowalter, line 47: what was the behavior before this?

packages/react-inbox/src/components/Messages2.0/Header.tsx
rileylnapier: it was rendering a 0 vs not rendering anything

packages/react-inbox/src/components/Messages2.0/Message.tsx
rileylnapier: if a brand can't be loaded it seems to render black so this is just a fallback

packages/react-provider/src/hooks/use-transport.ts
rileylnapier: i got feedback that we shouldn't throw

packages/react-inbox/src/components/Messages2.0/Header.tsx
joshshowalter: that's v odd. `unreadMessageCount` being 0 should evaluate to falsy. Does the `MarkAllRead` component have some other value in it?

joshshowalter: jk https://stackoverflow.com/questions/53048037/react-showing-0-instead-of-nothing-with-short-circuit-conditional-component

joshshowalter (approved):

PR #527, opened 2024-01-18
allow empty userid/clientkey if authorization provided

jonath0n:
## Description

Allow for empty `userId` and `clientKey` if `authorization` provided. Currently if `authorization` is used to authenticate, `clientKey` and `userID` are not needed but will prevent component rendering if empty/null.

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

> `Fix [#1]()`


rileylnapier (approved):

rileylnapier:
i think we fixed this with another PR recently

PR #531, opened 2024-01-25
refactor(theming and darkmode): better themeing with basic css variab…

rileylnapier:
…les and dark mode support

## Description

theming and darkmode support:

image
image
image

## Type of change

## Related issues

> `Fix [#1]()`


packages/react-provider/src/index.tsx
```diff
@@ -49,33 +54,51 @@ export const registerMiddleware = _registerMiddleware;
...
-  tenantId,
   apiUrl,
   authorization,
   brand,
   brandId,
   children,
   clientKey,
   id,
+  tenantId,
+  theme: _theme,
```
joshshowalter, line 101: just curious on the rename to `_theme`?

packages/react-inbox/src/components/Messages2.0/Header.tsx
```diff
@@ -35,9 +35,9 @@ const Container = styled.div<{ view?: string }>(({ theme }) => {
       fontSize: 16,
       fontWeight: 700,
       height: "42px",
-      color: "rgb(36, 50, 75)",
-      backgroundColor: "#F2F6F9",
-      borderBottom: "1px solid rgb(222, 232, 240)",
+      backgroundColor: "var(--ci-background)",
```
joshshowalter, line 38: minor thing, this should probably be `background` instead of backgroundColor in the case that someone passes in shorthand css for `background`

packages/react-provider/src/index.tsx
rileylnapier: because i merge it with some defaults later and id rather call that "theme" than "mergedTheme" i guess

rileylnapier: and i do this pattern already with _middleware

bblatnick32 (approved):

PR #573, opened 2024-05-16
refactor(update docs and fix initial props): update docs and fix comp…

rileylnapier:
…onents to not render

## Description

memoize rendering components so context doesn't cause infinite renders

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-inbox/src/components/Inbox/index.tsx
+++ b/packages/react-inbox/src/components/Inbox/index.tsx
@@ -297,13 +297,16 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
 
 const InboxWrapper: React.FunctionComponent<InboxProps> = (props) => {
   const courierContext = useCourier<{ inbox: InboxProps }>();
+  const contextExists = Boolean(courierContext);
 
-  if (!courierContext) {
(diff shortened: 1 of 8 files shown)


packages/react-inbox/src/components/Inbox/index.tsx
```diff
@@ -297,13 +297,16 @@ const Inbox: React.FunctionComponent<InboxProps> = (props) => {
 
 const InboxWrapper: React.FunctionComponent<InboxProps> = (props) => {
   const courierContext = useCourier<{ inbox: InboxProps }>();
+  const contextExists = Boolean(courierContext);
 
-  if (!courierContext) {
-    console.warn("Inbox: Missing Courier Provider");
-    return null;
-  }
+  return useMemo(() => {
```
bblatnick32, line 302: @rileylnapier Does this need to be memo'd?

bblatnick32 (approved):

packages/react-inbox/src/components/Inbox/index.tsx
rileylnapier: creates infinite renders without because 'courierContext' changes a lot

PR #577, opened 2024-05-24
feat(better reconnect): check online status and reconnect when we com…

rileylnapier:
…e back online

## Description

check online status and reconnect when we come back online

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/transport/src/courier/index.ts
+++ b/packages/transport/src/courier/index.ts
@@ -1,39 +1,40 @@
 import { CourierWS } from "../ws";
 import { Transport } from "../base";
 import { Interceptor } from "@trycourier/core";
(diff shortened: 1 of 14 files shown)


packages/transport/src/courier/index.ts
```diff
@@ -33,8 +33,6 @@ export class CourierTransport extends Transport {
       userSignature: options.userSignature,
     });
 
-    this.ws.connect();
```
rileylnapier, line 36: we handle connecting in a useEffect inside useTransport now.

this means if someone uses the transport without the hook they will manually need to connect

drew-y (approved):

PR #619, opened 2025-02-11
chore: upgraded version for jwt-decode

varo2397:
## Description

Upgraded version for `jwt-decode`

## Type of change

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-hooks/src/use-transport.ts
+++ b/packages/react-hooks/src/use-transport.ts
@@ -1,6 +1,6 @@
 import { useMemo, useRef, useEffect } from "react";
 import { CourierTransport, TransportOptions } from "@trycourier/transport";
-import jwtDecode from "jwt-decode";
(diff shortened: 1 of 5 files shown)


packages/react-hooks/src/use-transport.ts
```diff
@@ -1,6 +1,6 @@
 import { useMemo, useRef, useEffect } from "react";
 import { CourierTransport, TransportOptions } from "@trycourier/transport";
-import jwtDecode from "jwt-decode";
+import { jwtDecode } from "jwt-decode";
```
noahnu, line 3: The jwt-decode library is missing from the react-hooks package.json

rileylnapier:
Hi did we need this upgrade?

rileylnapier:
wait i think we did upgrade these in a sperate PR

Issue #620, opened 2025-02-28
Snyk reporting vulnerability which introduced by @trycourier/client-graphql.


asadbekabduvoitov-nordra:
![Image](https://github.com/user-attachments/assets/27d0ff43-d550-4d69-8be5-9bda3e621815)

Our CI/CD pipeline is getting failed, after installing @trycourier/client-graphql package for more flexibility. But some packages which are introduced after installing @trycourier/client-graphql might have vulnerability issues per Snyk scan results. Is this related to package or is there any other fixes that can be applied by patching ?

asadbekabduvoitov-nordra:
@rileylnapier yeah please, It's crucial for our CI/CD. 

Im leaving links for Snyk, if it helps. 

https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116

Would so helpful, if this can be tackled asap. 🙏

PR #623, opened 2025-03-04
C-13214 chore(upgrade rimraf): old rimraf had a dependency that had a depende…

rileylnapier:
…ncy that has memory leaks

## Description

## Type of change

## Related issues

should fix https://github.com/trycourier/courier-react/issues/620


gerjunior (approved):
I couldn't find any relevant breaking changes from rimraf 3.0 -> 5.0 given our current uses of it, so LGTM

PR #638, opened 2025-04-17
fix: ensure preference experience page will show digest time in browser timezone

antoniobritto07:
## Description

- The idea of this PR is to make sure digest schedule date, into Notification Preferences page, will be reflecting the user browser timezone. 

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

> `Fix [#1]()`

### DIFF
--- a/packages/react-preferences/src/utils/format_digest.ts
+++ b/packages/react-preferences/src/utils/format_digest.ts
@@ -1,5 +1,6 @@
 import { DigestSchedule, RepeatOn } from "@trycourier/core";
 import format from "date-fns/format";
+import { format as formatTZ, utcToZonedTime } from "date-fns-tz";
 
 export const toUpperCaseFirstLetter = (str: string) => {
   return str.charAt(0).toUpperCase() + str.slice(1);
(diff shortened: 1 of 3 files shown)


packages/react-preferences/src/utils/format_digest.ts
```diff
@@ -64,4 +67,25 @@ const getWeekdaysRepeatOnString = (repeatOn: RepeatOn) => {
   return weekdays.join(", ");
 };
 
+function formatDateBrowserTimezone(dateStrUTC: string, tz: string) {
```
antoniobritto07, line 70: that's the logic to show the digest schedule dates considering the timezone of the user's browser. That's because we've been seeing some hiccups due to DST. And a way to bypass it was to simulate the UTC time of the schedule as if it were happening today, then format it using the current timezone offset — which reflects whether DST is active or not.

yarn.lock
```diff
@@ -8612,6 +8612,11 @@ data-urls@^2.0.0:
     whatwg-mimetype "^2.3.0"
     whatwg-url "^8.0.0"
 
+date-fns-tz@^2.0.0:
```
antoniobritto07, line 8615: date-fns-tz needed to be added to the project, at the same version we do use in frontend repo

mikemilla (approved):

PR #685, opened 2026-04-16
fix(client-graphql): use updatePreferenceV2 mutation

mikemilla:
## Summary

- Switch `updateRecipientPreferences` from the v1 `updatePreferences` mutation (returns `Void`) to `updatePreferenceV2` (returns `RecipientPreference`) so the client returns the updated preference data
- Allow `digestSchedule` to accept `null` to unset a user's digest schedule preference (falls back to topic default)
- Add e2e test for setting `digestSchedule` via `updateRecipientPreferences`
- Fix test auth (use JWT) and GraphQL endpoint URL

Made with [Cursor](https://cursor.com)


packages/client-graphql/src/preferences.ts
```diff
@@ -145,8 +145,10 @@ export const getDraftPreferencePage =
 
 const UPDATE_RECIPIENT_PREFERENCES = `
   mutation UpdateRecipientPreferences($id: String!, $preferences: PreferencesInput!, $accountId: String) {
-    updatePreferences(templateId: $id preferences: $preferences accountId: $accountId) {
+    updatePreferenceV2(templateId: $id preferences: $preferences accountId: $accountId) {
```
mikemilla, line 148: Here is the change to fix the customer issue.

scarney81 (approved):

